PR #9, opened 2018-11-20
Added implementation with parameter

fryette:


wcoder:
@fryette 
Could you please provide (here) an easy example of using this navigation call?

fryette:
@wcoder you can use it like...
_pageNavigationService.For<MyViewModel>().WithParam(x => x.CustomParameterFieldOrProperty, parameterValue).Navigate();

Softeq.XToolkit.WhiteLabel/Navigation/NavigateHelper.cs
```diff
@@ -0,0 +1,47 @@
+﻿using System;
+using System.Linq.Expressions;
+using System.Reflection;
+using Softeq.XToolkit.WhiteLabel.Extensions;
+using Softeq.XToolkit.WhiteLabel.Mvvm;
+
+namespace Softeq.XToolkit.WhiteLabel.Navigation
+{
+    public class NavigateHelper<TViewModel> where TViewModel : IViewModelBase
```
wcoder: Maybe rename to `Builder` or something else, because of Helper associated with static class.

PR #14, opened 2018-12-05
Fixes with viewlocator

fryette:


fryette:
Dependencies willl be updated after checking whitelabel template

PR #26, opened 2019-02-21
Feature/test navigation extension

fryette:


Softeq.XToolkit.WhiteLabel.iOS/ViewControllerBase.cs
```diff
@@ -26,6 +26,7 @@ protected internal ViewControllerBase(IntPtr handle) : base(handle)
         }
 
         public List<IViewControllerComponent> ControllerComponents { get; } = new List<IViewControllerComponent>();
+        public abstract string ScreensGroupName { get; }
```
vadimpylsky: don't like this field
I think we should find better solution.

Softeq.XToolkit.WhiteLabel.iOS/Navigation/StoryboardNavigation.cs
```diff
@@ -40,11 +40,53 @@ public void GoBack()
...
+            }
+            else
+            {
+                throw new System.Exception($"ViewController of group {groupName} was not found");
+            }
+        }
+
         public void NavigateToViewModel(ViewModelBase viewModelBase, bool clearBackStack,
-            IReadOnlyList<NavigationParameterModel> parameters)
+            IReadOnlyList<NavigationParameterModel> parameters, string screensGroupName)
```
vadimpylsky: string screensGroupName: don't like this parameter in this method

PR #27, opened 2019-02-25
Destructive dialogs added

Synthezator:


Softeq.XToolkit.WhiteLabel.iOS/Services/StoryboardDialogsService.cs
```diff
@@ -67,14 +67,19 @@ public async Task<TViewModel> ShowForViewModel<TViewModel>(OpenDialogOptions opt
             return result as TViewModel;
         }
 
-        public Task<bool> ShowDialogAsync(string title, string message, string okButtonText, string cancelButtonText = null)
+        public Task<bool> ShowDialogAsync(string title,
```
fryette: Should we add instead of bool, enum with 
```
public enum DialogType
{
    Normal,
    Warning,
    .....?
}
```

PR #58, opened 2019-05-26
Add Playground Solution

wcoder:
Main goals of the Playground:
- support actual infrastructure;
- self-documentation of WL;


fryette:
Nice to have build job. Please add it.

PR #73, opened 2019-05-31
Switch ph1 to master

fryette:


fryette:
@wcoder Could you please support Android implementation?

wcoder:
@fryette Could you provide a list of changes that need to support on the Android side?

fryette:
> @fryette Could you provide a list of changes that need to support on the Android side?



> @fryette Could you provide a list of changes that need to support on the Android side?

The main issue is Dialog service.

PR #86, opened 2019-06-12
Add service locator registration to white label base classes

AlexandraMash:


Softeq.XToolkit.WhiteLabel.Droid/MainApplicationBase.cs
```diff
@@ -76,10 +76,18 @@ protected virtual void StartScopeForIoc()
...
+            RegisterServiceLocator(containerBuilder);
             RegisterInternalServices(containerBuilder);
             Dependencies.IocContainer.StartScope(containerBuilder);
         }
 
+        protected virtual void RegisterServiceLocator(ContainerBuilder builder)
+        {
+            var serviceLocator = new IocContainer();
+            Dependencies.Initialize(serviceLocator);
+            builder.Singleton<IIocContainer>(c => serviceLocator);
```
fryette: Add builder.Singleton<IIocContainer>(c => serviceLocator).PreserveExistingDefaults(); To avoid breaking changes in existed projects

Softeq.XToolkit.WhiteLabel.Droid/MainApplicationBase.cs
```diff
@@ -76,10 +76,18 @@ protected virtual void StartScopeForIoc()
...
             ConfigureIoc(containerBuilder);
+            RegisterServiceLocator(containerBuilder);
             RegisterInternalServices(containerBuilder);
             Dependencies.IocContainer.StartScope(containerBuilder);
         }
 
+        protected virtual void RegisterServiceLocator(ContainerBuilder builder)
+        {
+            var serviceLocator = new IocContainer();
+            Dependencies.Initialize(serviceLocator);
```
fryette: Please fix `RegisterServiceLocator` method to prevent exception in the existed projects. If you override method and call base you will have an exception in `Dependencies.Initialize` method that `ServiceLocator` was already registered. Please do the following code only if `servicelocator` not registered

Softeq.XToolkit.WhiteLabel.Droid/MainApplicationBase.cs
AlexandraMash: Added + used value from dependencies (in case they were previously initialized)

Softeq.XToolkit.WhiteLabel.Droid/MainApplicationBase.cs
AlexandraMash: Updated

Softeq.XToolkit.WhiteLabel.iOS/AppDelegateBase.cs
```diff
@@ -73,11 +72,23 @@ protected virtual void StartScopeForIoc()
         {
             var containerBuilder = new ContainerBuilder();
             ConfigureIoc(containerBuilder);
+            RegisterServiceLocator(containerBuilder);
             RegisterInternalServices(containerBuilder);
 
             Dependencies.IocContainer.StartScope(containerBuilder);
         }
 
+        protected virtual void RegisterServiceLocator(ContainerBuilder builder)
```
fryette: I see completely code duplicates, It would be great to move it to platform independent site and just reuse it on Android and iOS. @AlexandraMash what do you think?

AlexandraMash: Yes, I was thinking about the same thing. I was not sure which component can be used for this purpose, I don't think any of the existing ones will suite. And further maybe we could separate the whole flow of setting up Ioc from application and app delegate to some common component cause they are quite similar and besides are partially repeated in AutoRegistration version?

wcoder:
After discussion with @AlexandraMash, @fryette was implemented Bootstrapper concept:
- WL Core have BootstrapperBase where declared IoC container and setup WL dependencies
- WL platforms have Ios/DroidBoostrapper classes extended by BootstrapperBase where declared platform-specific dependencies.

- Each application must use WL platform-specific bootstrappers for extending.

Playground project contains actual demo.

wcoder:
Wiki page: https://github.com/Softeq/XToolkit.WhiteLabel/wiki/Bootstrapper

PR #90, opened 2019-06-13
Added wrapper on ioc container

fryette:
The main purpose of this request is to provide easy to use the wrapper to help switch on various IoC containers in the future. Dry IoC will come

Also removed Autofac from all projects except whitelabel(.net standart)


Softeq.XToolkit.WhiteLabel/Bootstrapper/AutofacContainerBuilder.cs
```diff
@@ -0,0 +1,147 @@
...
+
+        public IConcreteRegistration PerDependency<T1>(Func<IContainer, object> func)
+        {
+            return new AutofacRegistration<T1>(_builder.Register(c => func.Invoke(c.Resolve<IContainer>())).As<T1>()
+                .InstancePerDependency());
+        }
+
+        public IConcreteRegistration Singleton<T1, T2>() where T1 : T2
+        {
+            return new AutofacRegistration<T1>(_builder.RegisterType<T1>().As<T2>().InstancePerLifetimeScope().SingleInstance());
```
AlexandraMash: Looks like InstancePerLifetimeScope().SingleInstance() will do the same as simply SingleInstance(), perhaps this should be simplified

fryette: fixed

PR #92, opened 2019-06-14
refactor nav helpers

sleushunou:


fryette:
> I think we should avoid `Helper` naming
> Also, I suggest renaming to `FluentNavigatorBase<T>`, `PageFluentNavigator<T>` etc.
> Based on the concept of [Fluent interface ](https://en.wikipedia.org/wiki/Fluent_interface)
> 
> @sleushunou @fryette what do you think?



> I think we should avoid `Helper` naming
> Also, I suggest renaming to `FluentNavigatorBase<T>`, `PageFluentNavigator<T>` etc.
> Based on the concept of [Fluent interface ](https://en.wikipedia.org/wiki/Fluent_interface)
> 
> @sleushunou @fryette what do you think?

Agreed. Could be renamed.

PR #93, opened 2019-06-14
switch from autofac to dryioc

fryette:
Switched from [Autofac](https://github.com/autofac/Autofac) to [DryIoc](https://github.com/dadhi/DryIoc) container implementation.


fryette:
> Looks good!
> 
> I agree with switch to the DryIoc, but we need to test on our real projects, do you have test results?
> 
> Also, I prefer to merge previous changes (#90) with Autofac before applying these changes (after squash merge we'll lose history changes)

Sure. That's why this pull request targeting to separate branch(not master)

PR #124, opened 2019-07-06
Feature/convert nullable

Synthezator:


Softeq.XToolkit.Bindings/BindingGeneric.cs
```diff
@@ -1581,6 +1581,30 @@ public void SetConvertBack(Func<TTarget, TSource> convertBack)
...
+                    if (notNullableFromType != null)
+                    {
+                        if (value == null)
+                        {
+                            return default(TTo);
+                        }
+                        notNullableValue = System.Convert.ChangeType(value, notNullableFromType);
+                    }
+
+                    var notNullableToType = Nullable.GetUnderlyingType(typeof(TTo)) ?? typeof(TTo);
```
wcoder: As I can see, this method can be simplified to only 2 last lines, could you please check?

Softeq.XToolkit.Common/Extensions/DateTimeExtensions.cs
```diff
@@ -29,6 +29,12 @@ public static DateTime LastDayOfWeek(this DateTime date)
             return lastDay;
         }
 
+        public static DateTime WithKind(this DateTime date, DateTimeKind kind)
```
wcoder: Could you please check `DateTime.SpecifyKind` method, looks like the same?

https://docs.microsoft.com/en-us/dotnet/api/system.datetime.specifykind?view=xamarinios-10.8

Softeq.XToolkit.Bindings/BindingGeneric.cs
fryette: you can use string.IsNullOrEmpty()

PR #187, opened 2019-10-04
Refactor Common Library

wcoder:
### Description of Change ###

Changed Xtoolkit.Common library structure. Refactor & Cleanup.

### Issues Resolved ### 

None

### API Changes ###

Moved to Common:
- `ObservableObject`
- `ConsoleLogManager`
- `ConsoleLogger`

Moved to WL:
- `PagingModel`

Renamed:
- `Common.Command` to `Common.Commands`
- `Common.WeakSubscription` to `Common.Weak`
- `ILocalCache` to `ICache`

New namespaces:
- `Converters`
- `Logger`
- `Tasks`
- `Timers`

### Platforms Affected ### 

- Core (all platforms)

Xtoolkit.Common library structure.

Before | After
-----|----
Screen Shot 2019-10-04 at 1 17 16 PM  | Screen Shot 2019-10-04 at 1 25 31 PM

### PR Checklist ###


Softeq.XToolkit.WhiteLabel/Softeq.XToolkit.WhiteLabel.csproj
```diff
@@ -14,5 +14,9 @@
     <ProjectReference Include="..\Softeq.XToolkit.Common\Softeq.XToolkit.Common.csproj" />
     <ProjectReference Include="..\Softeq.XToolkit.Permissions\Softeq.XToolkit.Permissions.csproj" />
   </ItemGroup>
+
```
wcoder: Remove

Softeq.XToolkit.Common/Converters/IConverter.cs
```diff
@@ -0,0 +1,31 @@
...
+// http://www.softeq.com
+
+namespace Softeq.XToolkit.Common.Converters
+{
+    /// <summary>
+    ///     Represents methods for converting one type to another and back.
+    /// </summary>
+    /// <typeparam name="TOut">The type of the target property.</typeparam>
+    /// <typeparam name="TIn">The type of the source property.</typeparam>
+    public interface IConverter<TOut, TIn>
```
TimofeyBurak: Maybe there's a way to separate one-way and both-ways converters? Cause now, in order to implement this interface and use it in bindings it is required to implement both conversion methods, which leads to a lot of empty or fake implementations of ConvertValueBack(). Mostly, because sometimes it is just impossible to convert value back properly (for instance, converter "collection-to-is-empty")

wcoder: Good point!
I think we can resolve it via base converter:

```cs
public abstract class ConverterBase<TOut, TIn> : IConverter<TOut, TIn>
{
    public abstract TOut ConvertValue(TIn value, object parameter = null, string language = null);

    public virtual TIn ConvertValueBack(TOut value, object parameter = null, string language = null)
    {
        throw new System.NotImplementedException();
    }
}
```

Use:

Basic example:
```cs
public class MySuperConverter : ConverterBase<string, int>
{
    public override string ConvertValue(int value, object parameter = null, string language = null)
    {
        // ...
    }
}
```

Full example:
```cs
public class MySuperConverter : ConverterBase<string, int>
{
    public override string ConvertValue(int value, object parameter = null, string language = null)
    {
        // ...
    }

    public override int ConvertValueBack(string value, object parameter = null, string language = null)
    {
        // ...
    }
}
```

@TimofeyBurak What do you think?

TimofeyBurak: Not really what I meant.. I was thinking about how NotImplementedException can be avoided at all, by dividing cases where you need only one-way conversion, and where you need two-ways conversion

Synthezator: @TimofeyBurak you will need to re-implement another interface in case you need to add back-convertion. But maybe it's ok

Synthezator: I'm fine with ConverterBase

PR #191, opened 2019-10-09
Refactor Common library tasks extensions.

wcoder:
### API Changes ###

Changed:
- `SafeTaskWrapper` => `FireAndForget`
- `WithTimeoutAsync ` throws `TimeoutException` when timeout incorrect.
- `WithTimeout` deprecated.
 
 Removed:
 - `SafeTaskWrapper`

Added:
- `WithCancellation`
- `WithLoggingErrors`
 
### Platforms Affected ### 

- Core (all platforms)

None

### PR Checklist ###


Softeq.XToolkit.Common/Extensions/TaskExtensions.cs
```diff
@@ -11,43 +11,79 @@ namespace Softeq.XToolkit.Common.Extensions
...
 
-            return default;
+            using (var timerCancellation = new CancellationTokenSource())
+            {
+                var timeoutTask = Task.Delay(timeout, timerCancellation.Token);
+                var firstCompletedTask = await Task.WhenAny(task, timeoutTask).ConfigureAwait(false);
+
+                if (firstCompletedTask == timeoutTask)
+                {
+                    throw new TimeoutException();
```
TimofeyBurak: There's a peculiarity with this case. If the original task will throw an exception - it will never be handled

Softeq.XToolkit.Common/Extensions/TaskExtensions.cs
```diff
@@ -11,43 +11,79 @@ namespace Softeq.XToolkit.Common.Extensions
...
+        ///     A task that completes with the result of the specified <paramref name="task"/> or
+        ///     faults with a <see cref="TimeoutException"/> if <paramref name="timeout"/> elapses first.
+        /// </returns>
+        public static async Task<T> WithTimeout<T>(this Task<T> task, TimeSpan timeout)
         {
-            task.ContinueWith(t => { LogException(t, logger); }, CancellationToken.None,
-                TaskContinuationOptions.NotOnRanToCompletion, TaskScheduler.Default);
-            return task;
+            await WithTimeout((Task) task, timeout).ConfigureAwait(false);
+            return task.GetAwaiter().GetResult();
```
TimofeyBurak: Can this task just be awaited here?

Softeq.XToolkit.Common/Extensions/TaskExtensions.cs
```diff
@@ -11,43 +11,79 @@ namespace Softeq.XToolkit.Common.Extensions
...
         }
 
-        public static Task SafeTaskWrapper(this Task task, ILogger logger = null)
+        /// <summary>
+        ///     Useful for fire-and-forget calls to async methods.
+        /// </summary>
+        /// <param name="task">The task.</param>
+        /// <param name="logger">Logger object.</param>
+        /// <returns>The task.</returns>
+        public static void FireAndForget(this Task task, ILogger logger = null)
```
TimofeyBurak: The name might be a bit confusing. The task is not actually fired here (it might have been fired a long ago). Here we just make sure that already fired task is handled properly. This name is better suited for Func<Task> actually (although we cant make an extension for it)

Softeq.XToolkit.Common/Extensions/TaskExtensions.cs
Synthezator: Do we need to handle exceptions if they occur after timeout? Is it dangerous to not handle them?

TimofeyBurak: Tasks with unhandled exceptions might cause the following:
"System.AggregateException: A Task's exception(s) were not observed either by Waiting on the Task or accessing its Exception property. As a result, the unobserved exception was rethrown by the finalizer thread."
It's currently more than hundred thousands of such exceptions in HBI's hockeyapp, and the count is rising. Although it's usually not causing the app to actually crash, it floods crash analytics quite a lot.
BTW: the same thing might be caused by exception, thrown in async void method

wcoder: Some words about API backward-compatibility:
- Old impl. of method `WithTimeout` stays and deprecated. (As I can see, don't use anywhere)
- A new method was renamed to `WithTimeoutAsync`.
- Throwing exception it's correct behavior for this method, instead of `null`.

Softeq.XToolkit.Common/Extensions/TaskExtensions.cs
wcoder: Not need, the task already has result.

Softeq.XToolkit.Common/Extensions/TaskExtensions.cs
wcoder: Reimplemented.

Softeq.XToolkit.Common/Extensions/TaskExtensions.cs
TimofeyBurak: I'm not saying throwing exception is not correct. My concern is what gonna happened with original task after exception? It's not completed, and it's abandoned. What if it throw exception?

Softeq.XToolkit.Common/Extensions/TaskExtensions.cs
TimofeyBurak: Exactly! That's why there's no difference in using await and .GetAwaiter().GetResult() - both will return result immediately. But .GetAwaiter().GetResult() is still a blocking synchronous code in async method, i'd say - avoid it at all costs

Softeq.XToolkit.Common/Extensions/TaskExtensions.cs
wcoder: I looked at these cases:
- original Task throws an exception before timeout -> consumer get original exception;
- throws TimeoutException & than original Task throws internally exception -> consumer get TimeoutException, original task will have Exception internally. (How to get task exception, depends on consumer code.)

Softeq.XToolkit.Common/Extensions/TaskExtensions.cs
wcoder: Avoid it at all costs, yes. But depends on cases.
In the current case `await` has the same calls under the hood.
The problem can be in case when `WithTimeoutAsync` method will be changed.

Softeq.XToolkit.Common/Extensions/TaskExtensions.cs
wcoder: I propose the following solution for the second case:
```cs
var result = task.WithLoggingErrors(logger)
                 .WithTimeoutAsync(timeout);
```

PR #203, opened 2019-10-23
Refactor bootstrapper & containerBuilder

wcoder:
### Description of Change ###

- Fixed services registration order.
- Internal refactoring.

**Important**
- When you register new instances of internal services, you need to use `IfRegistered.Replace` otherwise you will have runtime exception about duplication registration.

**Additional notes**
- **Lazy resolving** - Because we use DryIoc under the hood, our `IContainer` has the implicit ability for lazy, factory resolving ([DryIoc doc](https://bitbucket.org/dadhi/dryioc/wiki/Wrappers#markdown-header-predefined-wrappers), [Autofac doc](https://autofaccn.readthedocs.io/en/latest/resolve/relationships.html#delayed-instantiation-lazy-b))

### API Changes ###

Obsoleted:
- `IContainer.ResolveLazy<TService>` look above.

Changed:
- enum `IfRegistered.AppendNewImplementation` => `IfRegistered.AppendNew`

### Platforms Affected ### 

- Core (all platforms)
- iOS
- Android

None

Not applicable

### PR Checklist ###

### DIFF
--- a/Softeq.XToolkit.WhiteLabel/Bootstrapper/Containers/DryIocContainerBuilder.cs
+++ b/Softeq.XToolkit.WhiteLabel/Bootstrapper/Containers/DryIocContainerBuilder.cs
@@ -6,6 +6,7 @@ using System.Collections.Generic;
 using DryIoc;
 using Softeq.XToolkit.Common.Extensions;
 using Softeq.XToolkit.WhiteLabel.Bootstrapper.Abstract;
+using IDryContainer = DryIoc.IContainer;
 using IContainer = Softeq.XToolkit.WhiteLabel.Bootstrapper.Abstract.IContainer;
 
 namespace Softeq.XToolkit.WhiteLabel.Bootstrapper.Containers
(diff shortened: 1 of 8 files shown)


Softeq.XToolkit.WhiteLabel/Bootstrapper/Containers/DryIocContainerBuilder.cs
```diff
@@ -69,39 +74,40 @@ public void RegisterBuildCallback(Action<IContainer> action)
 
         public IContainer Build()
         {
-            Singleton<DryIocContainer, IContainer>();
+            var container = new DryIocContainerAdapter();
 
-            var container = _container.Resolve<IContainer>();
-            _container = _container.WithNoMoreRegistrationAllowed();
+            _dryContainer.RegisterInstance<IContainer>(container, IfAlreadyRegistered.Keep);
```
TimofeyBurak: I wander if something like this could be possible:

in constructor of DryIocContainerAdapter
```
public DryIocContainerAdapter(DryIoc.IContainer container)
{
    container.RegisterInstance<IContainer>(this, IfAlreadyRegistered.Keep);
    _container = container.WithNoMoreRegistrationAllowed();
}
```

so Initialize() method would become redundant and here it would be just:
`var container = new DryIocContainerAdapter(_dryContainer);`

wcoder: Thanks for the detailed review!
Looks good, but in this case, we can't afford that, because `DryIocContainerBuilder` needs the latest (locked) instance of` DryIoc.IContainer`, and I would like to keep the adapter as simple as possible.

PR #218, opened 2019-11-18
Add Nullable support for Push Notifications

wcoder:


Softeq.XToolkit.PushNotifications/PushNotificationsServiceBase.cs
```diff
@@ -154,7 +154,7 @@ private bool TryParsePushNotification(object pushNotification, out PushNotificat
             catch (Exception ex)
             {
                 PushNotificationsHandler.HandleInvalidPushNotification(ex, pushNotification);
-                result = null;
+                result = new PushNotificationModel();
             }
 
             return result != null;
```
AlexandraMash: Looks like we will return true (parsing successful) in case of exception here

Issue #226, opened 2020-01-02
[iOS] Alert dialog with options


wcoder:
**Is your feature request related to a problem? Please describe.**

Extend/rework `ActionsSheetControl` for support aler dialogs with options.

https://github.com/Softeq/XToolkit.WhiteLabel/blob/a9bc7b0db605d71c4872c23ec8fe7ced0fe09b54/Softeq.XToolkit.WhiteLabel.iOS/Controls/ActionsSheetControl.cs#L49-L50

**Describe the solution you'd like**

Easy: Export the property for customizing `controllerStyle` argument of internal AlertViewController.

Complex: Separate dialog builder (for share internal behaviors of ActionsheetController, StoryboardDialogsService)

PR #239, opened 2020-02-13
Dialogs service extended

Synthezator:
### Issues Resolved ### 

- fixes #

### API Changes ###

 
 None

### Platforms Affected ### 

- Core (all platforms)
- iOS
- Android

None

Not applicable

### PR Checklist ###

### DIFF
--- a/Softeq.XToolkit.WhiteLabel.Droid/Dialogs/DroidFragmentDialogService.cs
+++ b/Softeq.XToolkit.WhiteLabel.Droid/Dialogs/DroidFragmentDialogService.cs
@@ -1,30 +1,35 @@
 // Developed by Softeq Development Corporation
 // http://www.softeq.com
 
+using System;
 using System.Collections.Generic;
 using System.Threading.Tasks;
+using Android.App;
+using Android.Content;
+using Plugin.CurrentActivity;
+using Softeq.XToolkit.Bindings;
+using Softeq.XToolkit.Common.Commands;
 using Softeq.XToolkit.WhiteLabel.Bootstrapper.Abstract;
 using Softeq.XToolkit.WhiteLabel.Droid.Navigation;
 using Softeq.XToolkit.WhiteLabel.Extensions;
 using Softeq.XToolkit.WhiteLabel.Model;
+using Softeq.XToolkit.WhiteLabel.Mvvm;
 using Softeq.XToolkit.WhiteLabel.Navigation;
 using Softeq.XToolkit.WhiteLabel.Navigation.FluentNavigators;
+using Softeq.XToolkit.WhiteLabel.Threading;
 
 namespace Softeq.XToolkit.WhiteLabel.Droid.Dialogs
 {
     public class DroidFragmentDialogService : IDialogsService
     {
-        private readonly IAlertBuilder _alertBuilder;
         private readonly IContainer _iocContainer;
         private readonly IViewLocator _viewLocator;
 
         public DroidFragmentDialogService(
             IViewLocator viewLocator,
-            IAlertBuilder alertBuilder,
             IContainer iocContainer)
         {
             _viewLocator = viewLocator;
-            _alertBuilder = alertBuilder;
             _iocContainer = iocContainer;
         }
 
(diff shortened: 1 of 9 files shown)


Softeq.XToolkit.WhiteLabel.iOS/Services/StoryboardDialogsService.cs
```diff
@@ -13,6 +13,8 @@
 using Softeq.XToolkit.WhiteLabel.Threading;
 using Softeq.XToolkit.WhiteLabel.Extensions;
 using UIKit;
+using Softeq.XToolkit.WhiteLabel.Mvvm;
```
TimofeyBurak: Seems like usings are not sorted

Softeq.XToolkit.WhiteLabel.iOS/Services/StoryboardDialogsService.cs
```diff
@@ -72,6 +74,39 @@ public class StoryboardDialogsService : IDialogsService
             return dialogResult.Task;
         }
 
+        public Task<TEnum> ShowDialogAsync<TEnum>(string title,
+            string message,
+            Dictionary<TEnum, DialogOption> actions)
+            where TEnum : Enum
```
TimofeyBurak: Should we really restrict usage to only enums? Seems like any type can be used here

Softeq.XToolkit.WhiteLabel.iOS/Services/StoryboardDialogsService.cs
```diff
@@ -72,6 +74,39 @@ public class StoryboardDialogsService : IDialogsService
...
+        {
+            var dialogResult = new TaskCompletionSource<TEnum>();
+
+            Execute.BeginOnUIThread(() =>
```
TimofeyBurak: Implementation of this method is very much looks like implementation of the default ShowDialogAsync, but this method is more versatile. Maybe we should reuse it?
BTW, seems like default implementation of the ShowDialogAsync might be turned into extension in this case

Softeq.XToolkit.WhiteLabel.iOS/Services/StoryboardDialogsService.cs
Synthezator: I thought about that too, will remove it

Softeq.XToolkit.WhiteLabel.iOS/Services/StoryboardDialogsService.cs
wcoder: I agree with @TimofeyBurak, related to an issue #226

Softeq.XToolkit.WhiteLabel/Navigation/IDialogsService.cs
```diff
@@ -18,10 +18,18 @@ public interface IDialogsService
...
+        /// Probably this must be refactored to use radiobuttons.
+        /// 
+        /// On iOS there can be only one Cancel option (it will be in the bottom) and several others,
+        /// which are not sorted.
+        /// </summary>
+        Task<TResult> ShowDialogAsync<TResult>(string title,
             string message,
-            Dictionary<TEnum, DialogOption> actions)
-            where TEnum : Enum;
+            Dictionary<TResult, DialogOption> actions);
```
wcoder: I propose to go another way:

A separate class without use as a base for `CommandAction`:

```cs
class DialogOption<T>
{
    public T Key { get;set; }
    public string Title { get; set; }
    public CommandActionStyle CommandActionStyle { get; set; }
}
```

```suggestion
            IList<DialogOption<TResult>> actions
```

Softeq.XToolkit.WhiteLabel.Droid/Dialogs/DroidFragmentDialogService.cs
```diff
@@ -35,7 +40,59 @@ public class DroidFragmentDialogService : IDialogsService
...
+
+                foreach (var action in actions)
+                {
+                    EventHandler<DialogClickEventArgs> onOptionChosen = (o, e) =>
+                    {
+                        tcs.TrySetResult(action.Key);
+                        var alertDialog = (AlertDialog) o;
+                        alertDialog.Dismiss();
+                    };
+                    if (action.Value.CommandActionStyle == CommandActionStyle.Destructive)
```
wcoder: Can be replaced to `switch`

wcoder:
This PR be closed because not actual after the latest, merged PR.

PR #260, opened 2020-03-02
Refactor bootstrapper

wcoder:
### Description of Change ###

Export duplicated logic to find view models by platform-specific view types.

### API Changes ###

Added:
 - class `BootstrapperWithViewModelLookup` - incapsulate logic about registering ViewModels and re-configuring ViewLocator;
 - class `ViewModelFinderBase` - incapsulate logic about lookup ViewModel types from known assemblies;
 - class `DroidViewModelFinder` - Android-specific finder;
 - class `IosViewModelFinder` - iOS-specific finder;
 - class `ViewModelToViewMap` - type alias to ViewModel to View mapping; 
 - class `ViewControllerHelper` - iOS helper with methods for instantiate ViewControllers;
 - class `BootstrapperBase`
   - method `RegisterTypesFromAssemblies(IContainerBuilder builder, IList<Assembly> assemblies)` - for make additional search and registrations from cached assemblies.

Changed:
 - class `BootstrapperBase`
   - method signature `BuildContainer(IContainerBuilder builder, IList<Assembly> assemblies)` to `BuildContainer(IContainerBuilder builder)`

### Platforms Affected ### 

- Core (all platforms)
- iOS
- Android

None

Not applicable

### PR Checklist ###

- [ ] Check startup performance degradation

### DIFF
--- a/Softeq.XToolkit.WhiteLabel.Droid/DroidBootstrapper.cs
+++ b/Softeq.XToolkit.WhiteLabel.Droid/DroidBootstrapper.cs
@@ -29,18 +29,8 @@ namespace Softeq.XToolkit.WhiteLabel.Droid
             // navigation
             builder.Singleton<ActivityPageNavigationService, IPlatformNavigationService>(IfRegistered.Keep);
             builder.Singleton<BundleService, IBundleService>(IfRegistered.Keep);
-            builder.Singleton<DroidViewLocator>(IfRegistered.Keep);
+            builder.Singleton<DroidViewLocator, IViewLocator>(IfRegistered.Keep);
             builder.PerDependency<DroidFrameNavigationService, IFrameNavigationService>(IfRegistered.Keep);
-        }
-
-        protected override void RegisterViewLocator(IContainerBuilder builder, Dictionary<Type, Type> viewModelToViewMapping)
-        {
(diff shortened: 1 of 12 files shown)


Softeq.XToolkit.WhiteLabel.Droid/DroidBootstrapper.cs
```diff
@@ -46,31 +33,26 @@ protected override void RegisterInternalServices(IContainerBuilder builder)
...
-
-                if (!viewModelType.IsAbstract)
-                {
-                    viewModelToViewTypes.Add(viewModelType, type);
-                    builder.PerDependency(viewModelType, IfRegistered.Keep);
-                }
-            }
-
-            return viewModelToViewTypes;
+                var viewLocator = container.Resolve<DroidViewLocator>();
```
TimofeyBurak: Does our IoC-container support registration with parameters? If so, it would be much better to just register DroidViewLocator as IViewLocator, providing viewModelToViewMapping as custom parameter.
Current implementation seems confusing, as it requires specific type to be already registered in IoC-container beforehand

wcoder: Great idea! Done.

PR #272, opened 2020-03-23
Rework dialogs

### LINKED ISSUE #226: [iOS] Alert dialog with options

wcoder:
### Description

Support integration with:
- Acr.Dialogs
- Xamarin.Forms
- custom impls.

### Issues Resolved

- fixes #226  #239

### API Changes

Deprecated:

- `IDialogsService.ShowDialogAsync(title, message, ...)`
- `ActionsSheetControl`
- `IAlertBuilder`
- `DefaultAlertBuilder`

Added:

- `IDialogsService.ShowDialogAsync<T>(IDialogConfig<T>)` - high-level method to show any dialogue;
- `*AlertDialog`, `AlertDialogConfig` - classes related to alert dialog;
- `*ConfirmDialog`, `ConfirmDialogConfig` - classes related to confirmation dialog;
- `*ActionSheetDialog`, `ActionSheetDialogConfig` - classes related to action sheet dialog;

### Platforms Affected

- Core (all platforms)
- iOS
- Android

### Behavioral/Visual Changes

- iOS: No custom `TintColor` for ActionSheet dialog;
- Android: No custom AlertStyle directly in dialog; (use activity theme)
- ~~Android: No cancellation of dialog by click to the background; (need to discuss)~~

Cancellation dialogs via tap by background or back button:
- Alert, Confirm - no (consistent behavior for supported platforms)
- ActionSheet - yes, if CancellationButtonText was set

### PR Checklist

### Subtasks
- [x] Refactor `IDialogsService` implementations
- [x] Improve mechanism for extend implementations
- [x] Add samples for dialogs to Playgrounds


Softeq.XToolkit.WhiteLabel.iOS/Dialogs/AlertAction.cs
```diff
@@ -0,0 +1,38 @@
...
+        {
+            return new AlertAction(text, UIAlertActionStyle.Default, action);
+        }
+
+        public static AlertAction Cancel(string text, Action action)
+        {
+            return new AlertAction(text, UIAlertActionStyle.Cancel, action);
+        }
+
+        public static AlertAction Destructive(string text, Action action, bool isDestructive = true)
```
TimofeyBurak: Seems a bit strange to be able to call `Destructive()` method with `isDestructive = false`. Seems like this case is already covered with `Default()`

Softeq.XToolkit.WhiteLabel/Dialogs/AlertDialogConfig.cs
```diff
@@ -0,0 +1,33 @@
...
+// http://www.softeq.com
+
+namespace Softeq.XToolkit.WhiteLabel.Dialogs
+{
+    /// <summary>
+    ///     Presents an alert dialog to the application user with a single cancel button.
+    /// </summary>
+    public class AlertDialogConfig : IDialogConfig<object?>
+    {
+        public AlertDialogConfig(string title, string message, string cancelButtonText = "OK")
```
TimofeyBurak: Default values of string won't be really useful in localizable projects, and might even be harmful

Softeq.XToolkit.WhiteLabel.Droid/Dialogs/DroidAlertDialog.cs
```diff
@@ -0,0 +1,37 @@
...
+        {
+            var dialogResult = new TaskCompletionSource<object?>();
+
+            Execute.BeginOnUIThread(() =>
+            {
+                var builder = GetBuilder()
+                    .SetTitle(_config.Title)
+                    .SetMessage(_config.Message);
+
+                SetPositiveButton(builder, _config.CancelButtonText, dialogResult, null);
```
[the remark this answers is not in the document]
wcoder: yeees, I agree 👍

Softeq.XToolkit.WhiteLabel/Dialogs/AlertDialogConfig.cs
```diff
@@ -0,0 +1,33 @@
+// Developed by Softeq Development Corporation
+// http://www.softeq.com
+
+namespace Softeq.XToolkit.WhiteLabel.Dialogs
+{
+    /// <summary>
+    ///     Presents an alert dialog to the application user with a single cancel button.
+    /// </summary>
+    public class AlertDialogConfig : IDialogConfig<object?>
```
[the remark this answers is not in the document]
wcoder: Good question!

Alert and confirm dialogs have some differences:
- the main is type of result;
- also can have different behavior for cancellation;
- explicit declaration of action

samples/Playground/Playground.Droid/Views/Dialogs/DialogsPageActivity.cs
```diff
@@ -29,14 +33,23 @@ protected override void OnCreate(Bundle savedInstanceState)
             SetContentView(Resource.Layout.activity_dialogs);
 
             _openAlertButton = FindViewById<Button>(Resource.Id.activity_dialogs_alert_button);
+            _openConfirmButton = FindViewById<Button>(Resource.Id.activity_dialogs_confirm_button);
```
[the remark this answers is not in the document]
wcoder: Also a good question! :)

Some time ago we already discuss the question of using Cheeseknife.
But I forget about the result :( 

In my opinion, CheeseKnife is easy to use, but reflection overhead in runtime.

Can be better to use LayoutCodeBinding for these cases:
https://github.com/xamarin/xamarin-android/blob/master/Documentation/guides/LayoutCodeBehind.md

Softeq.XToolkit.WhiteLabel/Extensions/DialogExtensions.cs
```diff
@@ -0,0 +1,31 @@
...
+{
+    public static class DialogExtensions
+    {
+        /// <summary>
+        ///     Type-matching for return explicit result.
+        /// </summary>
+        /// <param name="dialog">An instance of <see cref="IDialog{TResult}"/>.</param>
+        /// <typeparam name="TResult">Type of dialog result.</typeparam>
+        /// <returns></returns>
+        public static Task<TResult> ShowAsync<TResult>(this IDialog dialog)
```
TimofeyBurak: Doesn't look safe at all. Why would we ever need to do that? Providing that IDialog is empty, it's basically like casting object to a specific type

wcoder: Yes, you right, it's casting the object to a specific type, but provides more safety and sugary (for using, extending).

The main cause of this solution its single call-point of any dialogs (IDialogsService).

samples/Playground/Playground.Droid/Views/Dialogs/DialogsPageActivity.cs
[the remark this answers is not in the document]
SavikPavel: Looks cool, we should give it a try)

Softeq.XToolkit.WhiteLabel/Extensions/DialogExtensions.cs
TimofeyBurak: Sorry, but still can't think of any use case for IDialog interface and this extension. It seems like an abstraction for types, that don't actually have common interface.

- If we create single instance of dialog - we can use it as-is, through IDialog<T> interface, no need to cast it to IDialog;
- If we need a collection of dialogs, for some reason, and they all have different return types - then yes, we would need to use IDialog as a collection base type in order to store all of them. But I can't think of how to use such collection.. and how this extension might be useful in this case;

@wcoder can you compile a small example of how it might be used, please?

wcoder: > all have different return types

Yes, sure:

#### With IDialog:
```cs
public Task<T> ShowDialogAsync<T>(IDialogConfig<T> config)
{
    return config switch
    {
        ConfirmDialogConfig c => new DroidConfirmDialog(c).ShowAsync<T>(),
		ActionSheetDialogConfig ac => new DroidActionSheetDialog(ac).ShowAsync<T>(),
        ...
        _ => throw new ArgumentException($"Type of dialog config ({config.GetType()}) not supported", nameof(config)),
    };
}
```
> `DroidConfirmDialog` > `IDialog<bool>` > `IDialog`


#### Alternative, without IDialog & extension:
```cs
public Task<T> ShowDialogAsync<T>(IDialogConfig<T> config)
{
    return config switch
    {
        ConfirmDialogConfig c => (new DroidConfirmDialog(c).ShowAsync() as Task<T>)!,  <-- when null, inconsistency ??
		ActionSheetDialogConfig ac => (new DroidActionSheetDialog(ac).ShowAsync() as Task<T>)!,
        ...
        _ => throw new ArgumentException($"Type of dialog config ({config.GetType()}) not supported", nameof(config)),
    };
}
```
> `DroidConfirmDialog` > `IDialog<bool>`


The main complexity is due to interpreting `T` and concrete type as `bool`, `string`, `MyCustomResultType` for dialog result.

TimofeyBurak: OK, I finally understand what is the problem here. We are trying to use common entry-point for all types of the dialogs with the following method
`Task<T> ShowDialogAsync<T>(IDialogConfig<T> config)`
This signature basically says that no matter what configuration implementation you provide - we will be able to handle it, as it implements contract of `IDialogConfig<T>` interface (which is also empty currently..) But in reality, we support only very limited set of the configurations, as they 1-to-1 connected to specific implementations of dialogs. I see following solutions here:

1. We leave implementation as-is, but our API should reflect that we could work with very limited set of Dialog configurations. Basically, instead of 
```
Task<T> ShowDialogAsync<T>(IDialogConfig<T> config)
```
we need to settle with
```
Task<object?> ShowAlertDialogAsync(AlertDialogConfig config)
Task<bool> ShowConfirmationDialogAsync(ConfirmDialogConfig config)
Task<string> ShowActionSheetAsync(ActionSheetDialogConfig config)
```
2. We might actually add additional level of abstraction and actually let IDialogConfig<T> implementations build _any_ dialog with _any_ result type, with few of implementations already ready to use (alerts, confirmations and action sheets). I made a draft in the https://github.com/Softeq/XToolkit.WhiteLabel/tree/feature/dialogs-unification branch, it might need some enhancements, but it seem to work. Please, take a look

wcoder: Cool, @TimofeyBurak thank you for your vision and contribution! 👍

I was thinking so, from the beginning - use something like `IDialogBuilder` for describing dialog. But in this case, it seems like a more complex solution for understanding, support, extending.

Current disadvantages that I see:
-- Config - can't be reuse, because concretely limited by IDialogBuilder
-- IDialogBuilder - limited in expanding
-- DialogStyle - limited in expanding
-- DroidDialog, DroidDialogBuilder - can't be reuse
-- IocDialog, DroidDialogBuilder - can't be reuse

Good point for discussion:
- `KeyValuePait<string, T>` - ActionSheet options (currently, was skipped by default to consistent integration ability )

TimofeyBurak: > can't be reuse
You mean in another dialog infrastructure, if we gonna implement one? Then yes. Otherwise, don't really understand how they cannot be reused, sorry.

In any case, it seems like we need to determine what kind of API we want to get first. If we gonna support specific types of dialogs - we need to show that through interface and expose separate method for each type, and don't try to squash them in one. Or, if we gonna support extension from outside (even with limited tools of customization) - then we need to rework object model and really implement versatile mechanism of dialogs configuration.
I'm not insisting on each of this solutions, but we really need to decide before moving further.

wcoder: @TimofeyBurak thank you for the good discussion!

The main points that I used to define API:

1. simple declaration of dialog
2. single place for executing in IDialogsService
3. easy to customize default dialogs
4. provide a mechanism to use another implementation of dialogs (XF, Acr)

Proposed solution:

1. simple Config only with properties, result type by the interface;
2. `Task<T> ShowDialogAsync<T>(IDialogConfig<T> config)`;
3. default implementations for alert, confirm, actionSheet dialogs (can be used as base and extended)
4. simple IDialog interface - can be used any implementations of dialogs, different views per platform

Also, I added an example to extend dialogs. Defines the custom dialog that looks the same on Android and iOS but declared differently.

wcoder:
@TimofeyBurak, after our rich discussion I applied changes. Could you please check?

Main changes:
- Removed IDialog, IConfigDialog
- Methods `ShowDialogAsync` was splitted

wcoder:
@TimofeyBurak, @SavikPavel thank you for the review! 🚀

PR #280, opened 2020-03-29
Implement versatile Hash helper based on generic values, instead of objects

TimofeyBurak:
### Description

Hash helper function implementation, which use generics instead of objects. A bit more complex, but does not cause boxing for value types

### API Changes

Added:
 - static IHashCalculator Hash.Get() method
 - interface IHashCalculator
```
{
    IHashCalculator Using<T>(T value);
    int Calculate();
}
```

Changed:
 - static int HashHelper.GetHashCode() marked as obsolete with justification

### Platforms Affected

- Core (all platforms)

### PR Checklist

### DIFF
--- a/Softeq.XToolkit.Common/Helpers/Hashing/Hash.cs
+++ b/Softeq.XToolkit.Common/Helpers/Hashing/Hash.cs
@@ -1,0 +1,13 @@
+// Developed by Softeq Development Corporation
+// http://www.softeq.com
+
+namespace Softeq.XToolkit.Common.Helpers.Hashing
+{
+    /// <summary>
+    /// Class helps to get a hash code for a number of objects combined
+    /// </summary>
+    public static class Hash
+    {
+        public static IHashCalculator Get() => new HashCalculator();
+    }
+}
--- a/Softeq.XToolkit.Common/Helpers/Hashing/IHashCalculator.cs
+++ b/Softeq.XToolkit.Common/Helpers/Hashing/IHashCalculator.cs
@@ -1,0 +1,25 @@
+// Developed by Softeq Development Corporation
+// http://www.softeq.com
+
+namespace Softeq.XToolkit.Common.Helpers.Hashing
+{
+    /// <summary>
+    /// Common interface for Hash function implementations
+    /// </summary>
+    public interface IHashCalculator
+    {
+        /// <summary>
+        ///     Adds value to use in hash code calculations.
+        /// </summary>
+        /// <returns>IHashCalculator reference to use in chained calls</returns>
+        /// <typeparam name="T">Generic type of the value</typeparam>
+        /// <param name="value">Value to use in hash code calculations</param>
+        IHashCalculator Using<T>(T value);
+
+        /// <summary>
+        ///     Calculates hash code using specified objects.
+        /// </summary>
+        /// <returns>The hash code.</returns>
+        int Calculate();
+    }
+}
--- a/Softeq.XToolkit.Common.Tests/Helpers/HashHelperTests/HashHelperDataProvider.cs
+++ b/Softeq.XToolkit.Common.Tests/Helpers/HashHelperTests/HashHelperDataProvider.cs
@@ -230,5 +230,37 @@ namespace Softeq.XToolkit.Common.Tests.Helpers.HashHelperTests
                 }; // mixed
             }
         }
+
+        public static IEnumerable<object[]> HashValueTypeArgumentsData
+        {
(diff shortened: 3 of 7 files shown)


Softeq.XToolkit.Common/Helpers/Hashing/Hash.cs
```diff
@@ -0,0 +1,13 @@
...
+// http://www.softeq.com
+
+namespace Softeq.XToolkit.Common.Helpers.Hashing
+{
+    /// <summary>
+    /// Class helps to get a hash code for a number of objects combined
+    /// </summary>
+    public static class Hash
+    {
+        public static IHashCalculator Get() => new HashCalculator();
```
wcoder: `Get()` can be renamed to `Default` and used as property

Softeq.XToolkit.Common/Helpers/Hashing/IHashCalculator.cs
```diff
@@ -0,0 +1,25 @@
+// Developed by Softeq Development Corporation
+// http://www.softeq.com
+
+namespace Softeq.XToolkit.Common.Helpers.Hashing
+{
+    /// <summary>
+    /// Common interface for Hash function implementations
+    /// </summary>
+    public interface IHashCalculator
```
wcoder: `IHashCalculator` can be renamed to `IHashCodeCalculator`, to make an accent on the responsibility of abstraction

Softeq.XToolkit.Common/Helpers/Hashing/IHashCalculator.cs
```diff
@@ -0,0 +1,25 @@
...
+    {
+        /// <summary>
+        ///     Adds value to use in hash code calculations.
+        /// </summary>
+        /// <returns>IHashCalculator reference to use in chained calls</returns>
+        /// <typeparam name="T">Generic type of the value</typeparam>
+        /// <param name="value">Value to use in hash code calculations</param>
+        IHashCalculator Using<T>(T value);
```
wcoder: `Using` can be named to `Add`

TimofeyBurak: The idea behind naming was to make it more fluent, similar to speech. For instance, "Calculate hashcode using x,y,z" transforms into:
```
Hash.Get().Using(x).Using(y).Using(z).Calculate();
```
Another alternative was to use "of" instead of "using": "Calculate hashcode of x,y,z"..

The only thing that does not really fit it is "Get()" method. But again, the idea was that using hash-function abstraction (`interface IHashCalculator`) allows us to prepare different implementations of hash-functions for user to choose: FNV, Shift-And-Xor, Bernstein and so on:
```
Hash.FNV().Using(x).Using(y).Using(z).Calculate();
Hash.ShiftAndXor().Using(x).Using(y).Using(z).Calculate();
```
The problem was, I didn't find the name of algorithm that we are using right now, so I decided to give it default name of `Get()` for now. But I guess it could be renamed to `Default()`. The only thing I'm not sure of is to make it property. It is basically a factory method, that creates new instance of IHashCodeCalculator on each call.

Softeq.XToolkit.Common/Helpers/Hashing/IHashCalculator.cs
TimofeyBurak: Agreed

Softeq.XToolkit.Common/Helpers/Hashing/Hash.cs
TimofeyBurak: Please look at this [comment](https://github.com/Softeq/XToolkit.WhiteLabel/pull/280#discussion_r401395475)

wcoder:
@TimofeyBurak why we can't use [System.HashCode.Combine](https://docs.microsoft.com/en-us/dotnet/api/system.hashcode.combine) instead?

@vadimpylsky, @SavikPavel what do you think?

SavikPavel:
> @TimofeyBurak can we use [System.HashCode.Combine](https://docs.microsoft.com/en-us/dotnet/api/system.hashcode.combine) instead of `HashHelper` ...?
> 
> @vadimpylsky, @SavikPavel what do you think?

Sounds perfect to me. HashCode.Combine looks like the right option to do that

TimofeyBurak:
Created another PR #283 
Seems like this one could be closed

wcoder:
Ok, @TimofeyBurak thank you for your work!

PR #291, opened 2020-04-09
Add some converters and controls to WhiteLabel.Forms

AlexandraMash:
### Description

Added some common controls and converters to WhiteLabel.Forms

### API Changes

Added:
 - ToggleButton
 - BooleanInverseConverter
 - EqualityToBooleanConverter
 - IsNumberNotZeroConverter
 - IsNumberZeroConverter
 - MultiplyConverter
 - StringToUpperCaseConverter
 
### Platforms Affected

- Xamarin.Forms Core (all platforms)

### PR Checklist

### DIFF
--- a/Softeq.XToolkit.WhiteLabel.Forms/Controls/ToggleButton.cs
+++ b/Softeq.XToolkit.WhiteLabel.Forms/Controls/ToggleButton.cs
@@ -1,0 +1,90 @@
+﻿// Developed by Softeq Development Corporation
+// http://www.softeq.com
+
+using System;
+using System.Runtime.CompilerServices;
+using Xamarin.Forms;
+
+namespace Softeq.XToolkit.WhiteLabel.Forms.Controls
+{
+    /// <summary>
+    ///     A button with On/Off states. State names for VisualStateManager are
+    ///     <see cref="ToggledOnState"/>, <see cref="ToggledOffState"/>.
+    /// </summary>
+    public class ToggleButton : Button
+    {
+        /// <summary>
+        ///     On state name for VisualStateManager.
+        /// </summary>
+        public const string ToggledOnState = "ToggledOn";
+
+        /// <summary>
+        ///     Off state name for VisualStateManager.
+        /// </summary>
+        public const string ToggledOffState = "ToggledOff";
+
+        public static BindableProperty IsToggledProperty = BindableProperty.Create(
+            nameof(IsToggled),
+            typeof(bool),
+            typeof(ToggleButton),
+            false);
+
+        public static BindableProperty IsToggleEnabledProperty = BindableProperty.Create(
+            nameof(IsToggleEnabled),
+            typeof(bool),
+            typeof(ToggleButton),
+            true);
+
+        public ToggleButton()
+        {
+            Clicked += OnClick;
+        }
+
+        /// <summary>
+        ///     Event for button state changes.
+        /// </summary>
+        public event EventHandler<ToggledEventArgs>? Toggled;
+
+        /// <summary>
+        ///     Gets or sets a value indicating whether button state is On.
+        /// </summary>
+        public bool IsToggled
+        {
+            get => (bool) GetValue(IsToggledProperty);
+            set => SetValue(IsToggledProperty, value);
+        }
+
+        /// <summary>
+        ///     Gets or sets a value indicating whether button can be toggled.
+        /// </summary>
+        public bool IsToggleEnabled
+        {
+            get => (bool) GetValue(IsToggleEnabledProperty);
+            set => SetValue(IsToggleEnabledProperty, value);
+        }
+
+        protected override void OnParentSet()
+        {
+            base.OnParentSet();
+            VisualStateManager.GoToState(this, IsToggled ? ToggledOnState : ToggledOffState);
+        }
+
+        protected override void OnPropertyChanged([CallerMemberName] string propertyName = null)
+        {
+            base.OnPropertyChanged(propertyName);
+            if (propertyName == IsToggledProperty.PropertyName)
+            {
+                Toggled?.Invoke(this, new ToggledEventArgs(IsToggled));
+                VisualStateManager.GoToState(this, IsToggled ? ToggledOnState : ToggledOffState);
+            }
+        }
+
+        private void OnClick(object sender, EventArgs e)
+        {
+            if (IsToggleEnabled)
+            {
+                IsToggled = !IsToggled;
+            }
+        }
+    }
+}
--- a/Softeq.XToolkit.WhiteLabel.Forms/Converters/EqualityToBooleanConverter.cs
+++ b/Softeq.XToolkit.WhiteLabel.Forms/Converters/EqualityToBooleanConverter.cs
@@ -10,6 +10,8 @@ namespace Softeq.XToolkit.WhiteLabel.Forms.Converters
     /// <summary>
     ///     Convert equality between value and parameter to boolean result:
     ///     returns true if value equals parameter, false otherwise.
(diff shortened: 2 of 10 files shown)


Softeq.XToolkit.WhiteLabel.Forms/Converters/IsNumberNotZeroConverter.cs
```diff
@@ -0,0 +1,27 @@
...
+using Xamarin.Forms;
+
+namespace Softeq.XToolkit.WhiteLabel.Forms.Converters
+{
+    /// <summary>
+    ///     Convert integer value to true if it is not equal to zero, false otherwise.
+    ///     Can be used to changed visibility of element based on list count.
+    ///     ConvertBack is not implemented.
+    /// </summary>
+    public class IsNumberNotZeroConverter : IValueConverter
```
wcoder, line 15: It can be joined with `IsNumberZeroConverter` with a parameter like 'reverse'.

Softeq.XToolkit.WhiteLabel.Forms/Converters/StringToUpperCaseConverter.cs
```diff
@@ -0,0 +1,25 @@
...
+namespace Softeq.XToolkit.WhiteLabel.Forms.Converters
+{
+    /// <summary>
+    ///     Converts a string to UpperCase string.
+    /// </summary>
+    public class StringToUpperCaseConverter : IValueConverter
+    {
+        public object Convert(object value, Type targetType, object parameter, CultureInfo culture)
+        {
+            return ((string) value).ToUpper();
```
wcoder, line 17: It can be better to return `string.Empty` when the value is `null`?

Softeq.XToolkit.WhiteLabel.Forms/Converters/StringToUpperCaseConverter.cs
```diff
@@ -0,0 +1,25 @@
...
+using System;
+using System.Globalization;
+using Xamarin.Forms;
+
+namespace Softeq.XToolkit.WhiteLabel.Forms.Converters
+{
+    /// <summary>
+    ///     Converts a string to UpperCase string.
+    /// </summary>
+    public class StringToUpperCaseConverter : IValueConverter
```
wcoder, line 13: Also can be added `ToUpper` converter separately

Softeq.XToolkit.WhiteLabel.Forms/Converters/EqualityToBooleanConverter.cs
```diff
@@ -0,0 +1,27 @@
...
+using Xamarin.Forms;
+
+namespace Softeq.XToolkit.WhiteLabel.Forms.Converters
+{
+    /// <summary>
+    ///     Convert equality between value and parameter to boolean result:
+    ///     returns true if value equals parameter, false otherwise.
+    ///     ConvertBack is not implemented.
+    /// </summary>
+    public class EqualityToBooleanConverter : IValueConverter
```
wcoder, line 15: Could you please help with a better understanding of which case it can be used?

Softeq.XToolkit.WhiteLabel.Forms/Converters/MultiplyConverter.cs
```diff
@@ -0,0 +1,29 @@
...
+using System;
+using System.Globalization;
+using Xamarin.Forms;
+
+namespace Softeq.XToolkit.WhiteLabel.Forms.Converters
+{
+    /// <summary>
+    ///     Converts integer value multiplying it by parameter.
+    /// </summary>
+    public class MultiplyConverter : IValueConverter
```
wcoder, line 13: Could you please help with a better understanding of which case it can be used?

Softeq.XToolkit.WhiteLabel.Forms/Controls/ToggleButton.cs
```diff
@@ -0,0 +1,90 @@
...
+using System.Runtime.CompilerServices;
+using Xamarin.Forms;
+
+namespace Softeq.XToolkit.WhiteLabel.Forms.Controls
+{
+    /// <summary>
+    ///     A button with On/Off states. State names for VisualStateManager are
+    ///     <see cref="ToggledOnState"/>, <see cref="ToggledOffState"/>.
+    /// </summary>
+    public class ToggleButton : Button
```
wcoder, line 14: It seems like `Behavior`. Have you tried this?

Behavior can be used for `Button`, `ImageButton` ...

Softeq.XToolkit.WhiteLabel.Forms/Converters/StringToUpperCaseConverter.cs
AlexandraMash: Renamed.
@wcoder Please, explain what do you mean by 

> Also can be added ToUpper converter separately

wcoder: It may also be useful to have an opposite converter (separately)

Softeq.XToolkit.WhiteLabel.Forms/Converters/EqualityToBooleanConverter.cs
AlexandraMash: Added comment

Softeq.XToolkit.WhiteLabel.Forms/Converters/MultiplyConverter.cs
AlexandraMash: Added comment, but I agree that its usage might be a bit specific and we might not want to add it. Please see the use case I described

Softeq.XToolkit.WhiteLabel.Forms/Controls/ToggleButton.cs
AlexandraMash: I will look into this. The idea was that you can provide any styles for On/Off states so it can actually be a button with only ImageSource that changes (kinda the same as ImageButton) - I was actually using it as a custom check box that way. I was also initially using it in a custom tab selection control (with only 1 selected at a time) where these buttons were created programmatically depending on the amount of tabs needed and that was controlling their states - I guess that's why it was a better idea to make it a button extension and not behavior. I will think of turning it into behavior.

Softeq.XToolkit.WhiteLabel.Forms/Converters/IsNumberZeroConverter.cs
```diff
@@ -8,20 +8,37 @@
...
+            {
+                return (bool) parameter;
+            }
+            else if (parameter is string)
+            {
+                return bool.Parse((string) parameter);
+            }
+
+            return true;
+        }
```
wcoder, line 42:
```suggestion
        private bool GetParameter(object parameter)
        {
            return parameter switch
            {
                bool boolean => boolean,
                string str => bool.Parse(str),
                _ => true
            };
        }
```

Softeq.XToolkit.WhiteLabel.Forms/Controls/ToggleButton.cs
AlexandraMash: @wcoder the problem with Behavior is that Button and ImageButton only have common accessible ancestor View which does not have Clicked event (ImageButton does not inherit from Button as I would expect and IButtonElement is not accessible) so I do not see how I can make this behavior shared for Button and ImageButton. I still can make it a behavior for Button but I'm not sure if this makes sense

wcoder: Agree

PR #292, opened 2020-04-09
Feature/forms generic root frame navigation

AlexandraMash:
# Description

Added a generic root ViewModel/Page for Xamarin.Forms frame navigation to avoid creating a new identical pair of viewModel + page for each frame stack that you have (for instance side menu with 5 items - 5 stacks).

### API Changes

Added:
 - RootFrameNavigationPageViewModel<T> class
 - string BuildRootFrameNavigationPageTypeName(string viewModelTypeName) to FormsViewLocator that can be overriden if  you do not want to use provided generic class for root frame navigation

Usage of RootFrameNavigationPageViewModel:
- Do not create an additional ViewModel and Page for root frame navigation page (like DrillRootPageViewModel and DrillRootPage in Playground samlpe)
- Register RootFrameNavigationPageViewModel<Your_first_page_in_stack_vm> in Bootstrapper
- Resolve RootFrameNavigationPageViewModel<Your_first_page_in_stack_vm> when needed (example - MasterDetailViewModelFactory.cs  in Playground sample)

### Platforms Affected

- Xamarin.Forms Core (all platforms)

### PR Checklist

- [ ] I have read the [CONTRIBUTING](https://github.com/Softeq/XToolkit.WhiteLabel/blob/master/.github/CONTRIBUTING.md) document
- [ ] My code follows the [code styles](https://github.com/Softeq/dotnet-guidelines)


Softeq.XToolkit.WhiteLabel.Forms/Navigation/RootFrameNavigationPage.cs
```diff
@@ -0,0 +1,11 @@
+﻿// Developed by Softeq Development Corporation
+// http://www.softeq.com
+
+using Xamarin.Forms;
+
+namespace Softeq.XToolkit.WhiteLabel.Forms.Navigation
+{
+    public class RootFrameNavigationPage : NavigationPage
```
wcoder, line 8: Looks than in this case application can't customize navigation root

Softeq.XToolkit.WhiteLabel/Mvvm/RootFrameNavigationPageViewModel.cs
```diff
@@ -0,0 +1,19 @@
+﻿// Developed by Softeq Development Corporation
+// http://www.softeq.com
+
+using Softeq.XToolkit.WhiteLabel.Navigation;
+
+namespace Softeq.XToolkit.WhiteLabel.Mvvm
+{
+    public class RootFrameNavigationPageViewModel<T> : RootFrameNavigationViewModelBase where T : IViewModelBase
```
wcoder, line 8: Navigation behavior looks the same as `DrillRootPageViewModel` (navigation stack for each root)

Softeq.XToolkit.WhiteLabel.Forms/Navigation/RootFrameNavigationPage.cs
AlexandraMash: We still have opportunity to create separate navigation page + view model or navigation page only and override BuildRootFrameNavigationPageTypeName in ViewLocator, but feels like in most cases application won't need that

Softeq.XToolkit.WhiteLabel/Mvvm/RootFrameNavigationPageViewModel.cs
AlexandraMash: Sure. Behavior is the same - we provide navigation page with first page in stack. The difference is that in case of DrillRootPageViewModel if you need 5 stacks (say 5 elements in your side/tab menu) you have to create 5 such root view models in your project that do exactly the same, only set different first pages (in addition to your 5 first screen view models). In case of RootFrameNavigationPageViewModel<T> you do not need to create these additional view models and navigation pages and you can use this generic class that performs the same logic.

wcoder: Agree!

Softeq.XToolkit.WhiteLabel.Forms/Navigation/RootFrameNavigationPage.cs
wcoder: Agree!

PR #304, opened 2020-04-24
Improve/connectivity

wcoder:
# Description

- Refactor connectivity iOS
- Update connectivity documentation

### API Changes

 None

### Platforms Affected

- iOS

### PR Checklist

### DIFF
--- a/Softeq.XToolkit.Connectivity.iOS/IosConnectivityService.cs
+++ b/Softeq.XToolkit.Connectivity.iOS/IosConnectivityService.cs
@@ -36,7 +36,7 @@ namespace Softeq.XToolkit.Connectivity.iOS
             Dispose(false);
         }
 
-        public bool IsConnected => _connectionStatuses.Values.Any(x => x);
+        public bool IsConnected => CheckConnectivity(_connectionStatuses);
 
         public bool IsSupported => true;
 
@@ -67,6 +67,13 @@ namespace Softeq.XToolkit.Connectivity.iOS
 
                 _monitors.Clear();
             }
+        }
+
+        protected virtual bool CheckConnectivity(IReadOnlyDictionary<NWInterfaceType, bool> connectionStatuses)
+        {
+            return connectionStatuses
+                .Where(x => x.Key == NWInterfaceType.Wifi || x.Key == NWInterfaceType.Cellular)
+                .Any(x => x.Value);
         }
 
         protected virtual IEnumerable<ConnectionType> FilterConnectionTypes(IList<NWInterfaceType> activeNetworkTypes)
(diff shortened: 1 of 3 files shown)


documentation/articles/xtoolkit/connectivity.md
```diff
@@ -6,6 +6,28 @@ Library over the [Xam.Plugin.Connectivity](https://github.com/jamesmontemagno/Co
 
 Currently can be used only as source code (via git submodules).
 
+## Setup
+
+### Project Reference
+
+Add references to **Softeq.XToolkit.Connectivity.*** for projects which you need to use **Connectivity** component.
```
SavikPavel, line 13: Maybe it's better to use something like that?
```suggestion
Add references to **Softeq.XToolkit.Connectivity.*** for projects where you need to use **Connectivity** component.
```

Softeq.XToolkit.Connectivity.iOS/IosConnectivityService.cs
```diff
@@ -64,47 +55,55 @@ public void Dispose()
...
-            HandleUpdateSnapshot(nWPath, NWInterfaceType.Loopback);
+            _connectionStatuses.TryAdd(type, false);
+
+            var monitor = CreateMonitor(type, nWPath => HandleUpdateSnapshot(nWPath, type));
+            monitor.Start();
+            return monitor;
         }
 
-        private void UpdateOtherSnapshot(NWPath nWPath)
+        private NWPathMonitor CreateMonitor(NWInterfaceType type, Action<NWPath> action)
```
SavikPavel, line 101: Looks like this method is only used in `RegisterMonitor.` Maybe it makes sense to rename `RegisterMonitor` to something like `CreateAndRegisterMonitor` and get rid of `CreateMonitor` method completely?
I'm not really sure this will be better than it is right now, but this is just an idea. What do you think?

documentation/articles/xtoolkit/connectivity.md
```diff
@@ -16,6 +38,15 @@ Please, use [IConnectivityService](xref:Softeq.XToolkit.Connectivity.IConnectivi
 var isConnected = _connectivityService.IsConnected;
 ```
 
+**Advanced**: Strict check connectivity only for most common mobile connections:
```
SavikPavel, line 41: Can you please point out what is the exact problem that is fixed by this code? It's not really obvious when you should use a simple connectivity check and when you should use the solution you've provided here

wcoder: Deleted

Softeq.XToolkit.Connectivity.iOS/IosConnectivityService.cs
wcoder: `CreateMonitor` only creates the monitor, one responsibility (can be exported in the future as utility).

`RegisterMonitor` more complex method, describes the specific features of registration.

I think we should leave the methods separately.

PR #317, opened 2020-07-14
Rework Bootstrapper responsibility and application initialization lifecycle

TimofeyBurak:
### Description

Make IBootstrapper stateless IoC container builder. Add callback with initialized container for AppDelegate and MainApplication

### API Changes

Added:
 - `void virtual MainApplicationBase.OnContainerInitialized(IContainer)`;
 - `void virtual AppDelegateBase.OnContainerInitialized(IContainer)`;

Changed:
 - `void IBootstrapper.Initialize()` => `IContainer IBootstrapper.Initialize()`;
 - `IBootstrapper MainApplicationBase.Bootstrapper {get;}` => `IBootstrapper MainApplicationBase.CreateBootstrapper()`;
 - `IBootstrapper AppDelegateBase.Bootstrapper {get;}` => `IBootstrapper AppDelegateBase.CreateBootstrapper()`;

### Platforms Affected

- Core (all platforms)
- iOS
- Android

### Behavioral/Visual Changes

None

### Before/After Screenshots

Not applicable

### PR Checklist

### DIFF
--- a/Softeq.XToolkit.WhiteLabel.Droid/MainApplicationBase.cs
+++ b/Softeq.XToolkit.WhiteLabel.Droid/MainApplicationBase.cs
@@ -9,6 +9,7 @@ using Plugin.CurrentActivity;
 using Softeq.XToolkit.Bindings;
 using Softeq.XToolkit.Bindings.Droid;
 using Softeq.XToolkit.WhiteLabel.Bootstrapper;
+using Softeq.XToolkit.WhiteLabel.Bootstrapper.Abstract;
 using Softeq.XToolkit.WhiteLabel.Droid.Providers;
 using Softeq.XToolkit.WhiteLabel.Threading;
 
@@ -24,8 +25,6 @@ namespace Softeq.XToolkit.WhiteLabel.Droid
             : base(handle, transfer)
         {
         }
-
-        protected abstract IBootstrapper Bootstrapper { get; }
 
         public override void OnCreate()
         {
@@ -66,6 +65,8 @@ namespace Softeq.XToolkit.WhiteLabel.Droid
             CrossCurrentActivity.Current.Init(this);
         }
 
+        protected abstract IBootstrapper CreateBootstrapper();
+
         protected virtual void InitializeWhiteLabelRuntime()
         {
             // Init Bindings
@@ -75,7 +76,16 @@ namespace Softeq.XToolkit.WhiteLabel.Droid
             PlatformProvider.Current = new DroidPlatformProvider();
 
             // Init dependencies
-            Bootstrapper.Initialize();
+            var bootstrapper = CreateBootstrapper();
+            var container = bootstrapper.Initialize();
+            Dependencies.Initialize(container);
+
+            // Notify dependencies ready to be used
+            OnContainerInitialized(container);
+        }
+
+        protected virtual void OnContainerInitialized(IContainer container)
+        {
         }
     }
 }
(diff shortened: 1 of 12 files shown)


Softeq.XToolkit.WhiteLabel.iOS/AppDelegateBase.cs
```diff
@@ -17,15 +18,15 @@ public abstract class AppDelegateBase : UIApplicationDelegate
...
-        protected abstract IBootstrapper Bootstrapper { get; }
-
         public override bool FinishedLaunching(UIApplication application, NSDictionary launchOptions)
         {
             InitializeWhiteLabelRuntime();
 
             return true;
         }
 
+        protected abstract IBootstrapper CreateBootstrapper();
```
wcoder, line 28: May be better to use as a property
Point to discuss with the team.

Softeq.XToolkit.WhiteLabel.iOS/AppDelegateBase.cs
```diff
@@ -35,7 +36,16 @@ protected virtual void InitializeWhiteLabelRuntime()
             PlatformProvider.Current = new IosPlatformProvider();
 
             // Init dependencies
-            Bootstrapper.Initialize();
+            var bootstrapper = CreateBootstrapper();
+            var container = bootstrapper.Initialize();
+            Dependencies.Initialize(container);
```
wcoder, line 41: Good for Bootstrapper, but can be not so obvious at all for infrastructure.

Softeq.XToolkit.WhiteLabel.Droid/MainApplicationBase.cs
```diff
@@ -75,7 +76,16 @@ protected virtual void InitializeWhiteLabelRuntime()
             PlatformProvider.Current = new DroidPlatformProvider();
 
             // Init dependencies
-            Bootstrapper.Initialize();
+            var bootstrapper = CreateBootstrapper();
+            var container = bootstrapper.Initialize();
+            Dependencies.Initialize(container);
```
wcoder, line 81: Also, need to add `Dependencies.Initialize(container);` to Xamarin.Forms:
https://github.com/Softeq/XToolkit.WhiteLabel/blob/master/Softeq.XToolkit.WhiteLabel.Forms/FormsApp.cs

TimofeyBurak: Indeed, done. I've also updated OnStarted() callback to accept configured container

Softeq.XToolkit.WhiteLabel.iOS/AppDelegateBase.cs
TimofeyBurak: As I understand, we don't really need to keep an instance of the Bootstrapper once it is initialized. And if it is kept - it is possible to use it second time (I believe that's why there was a check for "isInitialized" [here](https://github.com/Softeq/XToolkit.WhiteLabel/pull/317/files#diff-1584b30f91f5c1bf8f3953bc498d73b3L25))

Softeq.XToolkit.WhiteLabel.iOS/AppDelegateBase.cs
TimofeyBurak: To my understanding, the only responsibility of Bootstrapper is to configure IoC container (now it should be obvious just by looking into IBootstrapper interface). So I moved registration of the container in the app infrastructure outside - to AppDelegete and MainApplication respectively. There's even a method there `InitializeWhiteLabelRuntime()` - so AppDelegete and MainApplication are really responsible for initializing a runtime, and IoC container is just a part of it.
And one last point: AppDelegete and MainApplication expect an implementation of IBootstrapper interface, theoretically it could be any implementation, and it is not obvious that this implementation **must** call `Dependencies.Initialize(container);`

wcoder: Cool, I agree.

Softeq.XToolkit.WhiteLabel.iOS/AppDelegateBase.cs
wcoder: I agree.

PR #319, opened 2020-07-27
Add custom borders with support AutoLayout

wcoder:
### Description

Ability to add a border to the custom edges:

```cs
testView
    .WithBorder(8, UIColor.Red.CGColor, UIRectEdge.Left)
    .WithBorder(8, UIColor.Blue.CGColor, UIRectEdge.Top | UIRectEdge.Bottom);
```

### API Changes

Added:
 - UIView UIView.WithBorder(this UIView view, nfloat borderWidth, CGColor borderColor, UIRectEdge edge)

### Platforms Affected

- iOS

### PR Checklist

### DIFF
--- a/Softeq.XToolkit.Common.iOS/Extensions/UIViewExtensions.cs
+++ b/Softeq.XToolkit.Common.iOS/Extensions/UIViewExtensions.cs
@@ -47,47 +47,45 @@ namespace Softeq.XToolkit.Common.iOS.Extensions
 
             if (edge.HasFlag(UIRectEdge.Top))
             {
-                AddBorder(view, borderWidth, borderColor, UIRectEdge.Top);
+                AddBorder(
+                    view,
+                    borderColor,
+                    UIViewAutoresizing.FlexibleWidth | UIViewAutoresizing.FlexibleBottomMargin,
+                    new CGRect(0, 0, view.Frame.Size.Width, borderWidth));
             }
 
             if (edge.HasFlag(UIRectEdge.Bottom))
             {
-                AddBorder(view, borderWidth, borderColor, UIRectEdge.Bottom);
+                AddBorder(
+                    view,
+                    borderColor,
+                    UIViewAutoresizing.FlexibleWidth | UIViewAutoresizing.FlexibleTopMargin,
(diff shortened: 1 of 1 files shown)


Softeq.XToolkit.Common.iOS/Extensions/UIViewExtensions.cs
```diff
@@ -25,6 +25,87 @@ public static UIView WithBorder(this UIView view, nfloat borderWidth, CGColor? b
...
+                AddBorder(view, borderWidth, borderColor, UIRectEdge.Right);
+            }
+
+            return view;
+        }
+
+        private static UIView AddBorder(UIView view, nfloat borderWidth, CGColor borderColor, UIRectEdge edge)
+        {
+#pragma warning disable CS8509 // The switch expression does not handle all possible values of its input type (it is not exhaustive).
+            var autoresizingMask = edge switch
```
TimofeyBurak, line 74: Maybe we could pass `autoresizingMask` and `frame` from outside instead of switching, as we should already know what type of border we are trying to add? For instance
```
            if (edge.HasFlag(UIRectEdge.Right))
            {
                AddBorder(
                    view, 
                    borderWidth, 
                    borderColor, 
                    UIViewAutoresizing.FlexibleHeight | UIViewAutoresizing.FlexibleLeftMargin,
                    new CGRect(view.Frame.Size.Width - borderWidth, 0, borderWidth, view.Frame.Size.Height));
            }
```
or enumerate through flags in WithBorder(..., UIRectEdge edge) method like this:
```
            foreach (var flag in edge.GetFlags())
            {
                AddBorder(view, borderWidth, borderColor, flag);
            }
            
            where
            
            public static IEnumerable<Enum> GetFlags(this Enum e)
            {
                return Enum.GetValues(e.GetType()).Cast<Enum>().Where(e.HasFlag);
            }
```

wcoder: Passing `autoresizingMask` and `frame` good idea
Done.

foreach can't be simply applied because we need different ways for some elements

PR #320, opened 2020-07-27
Add Validation

### LINKED ISSUE #144: [Draft] Validation Spec
### Spec
```cs
/// <summary>
/// Provides a way for an object to be validated.
/// </summary>
/// <typeparam name="T">Type of the data to be validated</typeparam>
interface IValidatable
{
    /// <summary>
    /// The value indicating whether the validation succeeded.
    /// </summary>
    bool IsValid { get; }

    /// <summary>
    /// Used to  perform the validations of the property
    /// </summary>
    /// <returns>Gets a value indicating whether the validation succeeded.</returns> 
    bool Validate();
}
```
```cs
class ValidatableObject<T> : ObservableObject, IValidatable
{
    /// ...

    /// <summary>
    /// Reference value is attempting to assign to the property.
    /// </summary>
    T Value { get; }
    
    /// <summary>
    /// List of standards or validations you specify before the user can save the record
    /// </summary>
    IList<IValidator> Validators { get; }

    /// <summary>
    /// List of errors users have before can save the record
    /// </summary>
    IList<string> Errors { get; }
}
```

Proposal to depends on [FluentValidation](https://github.com/JeremySkinner/FluentValidation)

wcoder:
### Description

Added basic validation approach.

### Issues Resolved

- fixes #144

### API Changes

Added:

- XToolkit.WhiteLabel.Validation
  - IValidationRule
  - IValidatableObject
  - ValidatableObject
  - ValidatableGroup
  - Rules
    - EmailRule
    - IsNotNullOrEmptyRule
    - IsValueTrueRule
    - LengthRule

### Platforms Affected

- Core (all platforms)

### PR Checklist

### Tasks

- [x] Samples
- [x] Tests
- [x] Documentation


Softeq.XToolkit.WhiteLabel/Validation/Rules/LengthRule.cs
```diff
@@ -0,0 +1,38 @@
+// Developed by Softeq Development Corporation
+// http://www.softeq.com
+
+using System;
+
+namespace Softeq.XToolkit.WhiteLabel.Validation.Rules
+{
+    public class LengthRule : IValidationRule<string>
```
SavikPavel, line 8: Maybe it would be better to name it "StringLengthRule" or something like that? (to mark, that this will only work with Strings)

Softeq.XToolkit.WhiteLabel/Validation/Rules/IsNotNullOrEmptyRule.cs
```diff
@@ -0,0 +1,20 @@
+// Developed by Softeq Development Corporation
+// http://www.softeq.com
+
+namespace Softeq.XToolkit.WhiteLabel.Validation.Rules
+{
+    public class IsNotNullOrEmptyRule : IValidationRule<string>
```
SavikPavel, line 6: Maybe it would be better to name it "IsStringEmpty" or something like that? (to mark, that this will only work with Strings). Also, it's strange to see "IsNotNullOrEmpty" in class name and "IsNullOrWhiteSpace" inside the Check method

wcoder: Renamed to `StringNotEmptyRule`

Softeq.XToolkit.WhiteLabel/Validation/ValidatableObject.cs
```diff
@@ -0,0 +1,80 @@
...
+            }
+        }
+
+        public bool IsValid
+        {
+            get => _isValid;
+            private set => Set(ref _isValid, value);
+        }
+
+        public ObservableRangeCollection<string> Errors { get; }
```
TimofeyBurak, line 49: Is it intended that Errors is may be modifiable from outside? Seems like it should be controlled by internal logic only

wcoder: ValidableObject control only internal validators, errors collection is a validators output, and can be changed externally after for example any server validators response and etc.)

Softeq.XToolkit.WhiteLabel/Validation/Rules/EmailRule.cs
```diff
@@ -0,0 +1,30 @@
...
+        public string ValidationMessage { get; }
+
+        public bool Check(string value)
+        {
+            if (string.IsNullOrEmpty(value))
+            {
+                return false;
+            }
+
+            var regex = new Regex(@"^([\w\.\-]+)@([\w\-]+)((\.(\w){2,3})+)$");
```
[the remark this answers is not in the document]
wcoder: Replaced with another default pattern. (from emailregex.com)

Also, now can be used different patterns from:
- [docs](https://docs.microsoft.com/en-us/dotnet/standard/base-types/how-to-verify-that-strings-are-in-valid-email-format?redirectedfrom=MSDN)
- [.net sources](https://github.com/Microsoft/referencesource/blob/master/System.ComponentModel.DataAnnotations/DataAnnotations/EmailAddressAttribute.cs)
- [fluent validator - email](https://github.com/FluentValidation/FluentValidation/blob/265fef0aed50183a8a48be4ffc0abd40c2ccaa2e/src/FluentValidation/Validators/EmailValidator.cs#L45)

Softeq.XToolkit.WhiteLabel/Validation/ValidatableObject.cs
TimofeyBurak: For this case, maybe it is better to add some kind of `AddError(string)` method to `ValidatableObject`?

wcoder: I added `AddError(string)` previously and revert changes, because in this case, we have several ways to add errors. Also, we have cases when needs reset errors from external.

TimofeyBurak: We may add `Clear()` method for that) My idea is that class should explicitly describe it capabilities through interface. In my opinion, it is not really obvious that you need to manipulate the collection in order to achieve certain behavior. Without looking into source code, at least.
Another idea, which is definitely more complicated, but probably more correct in terms of abstraction model - is to add support for async validation rules.

wcoder: I agree to add `ClearErrors()` and `AddError()` but in this case, we still will have two approaches to use errors.

Async rules approach not so often used in previous projects. Also, need another interface and implementation, can be overhead for now.
@TimofeyBurak Maybe do you have an example of async rules?

wcoder: @TimofeyBurak I think we already have a public interface for manage ValidatableObject -> Errors -> Add / Clear
case ValidatableObject -> AddError would be redundant for object

Softeq.XToolkit.WhiteLabel.Tests/Validation/Rules/EmailRuleTests.cs
```diff
@@ -0,0 +1,58 @@
...
+        }
+
+        [Fact]
+        public void Check_CustomEmailPattern_Valid()
+        {
+            var rule = new EmailRule("error message", @"^([\w\.\-]+)@([\w\-]+)((\.(\w){2,3})+)$");
+
+            var isValid = rule.Check("test@domain.online");
+
+            Assert.False(isValid);
```
AlexandraMash, line 55: The name of the test suggests that it should be true(valid)

AlexandraMash (changes requested):

Softeq.XToolkit.WhiteLabel.Tests/Validation/Rules/EmailRuleTests.cs
wcoder: renamed

AlexandraMash (approved):

wcoder:
Any change requests can be made using separate PRs.

PR #354, opened 2020-09-04
Feature/observable range collection tests

SavikPavel:
### Description

Tests for ObservableRangeCollection

### API Changes
 
 None

### Platforms Affected

- Core (all platforms)

### Behavioral/Visual Changes

None

### Before/After Screenshots

Not applicable

### PR Checklist


Softeq.XToolkit.Common/Collections/ObservableRangeCollection.cs
```diff
@@ -211,22 +211,19 @@ public void InsertRange(IEnumerable<T> collection, int startIndex)
...
-            for (var i = 0; i < changedItems.Count; i++)
+            var removedItems = new List<T>();
+            foreach (var item in collection)
             {
-                if (!Items.Remove(changedItems[i]))
+                if (Items.Remove(item))
                 {
-                    changedItems.RemoveAt(i); // Can't use a foreach because changedItems is intended to be (carefully) modified
-                    i--;
+                    removedItems.Add(item);
```
wcoder, line 222: Broken the previous comment.
Enumerable collection can be changed.

SavikPavel: As I understand, the logic was the following:
1. `collection` list is passed as a param of the RemoveRange method
2. `changedItems` list is initialized with the `collection` items
3. Then for each item of the `changedItems` list, we're trying to remove it from our ObservableRangeCollection, and if it's NOT removed, we're removing it from the `changedItems` list
4. `changedItems` list is changed while enumerating so that's why we couldn't use `foreach` here

Right now the logic has changed to initialize `changedItems` with empty collection and add item to it only if the item is removed from the ObservableRangeCollection. Also, we're enumerating through the original `collection` list, not through the `changedItems`.
As I understand, the problem that was mentioned in this comment is no longer actual

wcoder: Agree. Thanks for the detailed description!

Softeq.XToolkit.Common/Collections/ObservableRangeCollection.cs
```diff
@@ -211,22 +211,19 @@ public void InsertRange(IEnumerable<T> collection, int startIndex)
...
                 }
             }
 
             OnPropertyChanged(EventArgsCache.CountPropertyChanged);
             OnPropertyChanged(EventArgsCache.IndexerPropertyChanged);
             OnCollectionChanged(
-                new NotifyCollectionChangedEventArgs(NotifyCollectionChangedAction.Remove, changedItems, index));
+                new NotifyCollectionChangedEventArgs(NotifyCollectionChangedAction.Remove, removedItems));
```
wcoder, line 229: What about `startingIndex` of the event args?

SavikPavel: I was worried about it as well, but then I read the note in documentation to this method:
"NOTE: with notificationMode = Remove, removed items starting index is not set because items are not guaranteed to be consecutive."
Looks like this index was not valid anyways, so I've just removed it)

wcoder: Make sense, but currently, Bindings Collections used OldStartingIndex to handle collection changes.

The default value is `-1` does it affects our collection handlers for iOS and Android?

wcoder (approved):

PR #366, opened 2020-09-09
Rework uitextfield/view extensions

wcoder:
# Description

Usage:

```cs

var filter = new GroupFilter(
    new LengthFilter(3),
    new ForbiddenCharsFilter('#'),
    new AnyCustomFilter());

// UITextField
textField.SetFilter(filter);

// or UITextView
textView.SetFilter(filter);

// or UISearchBar

```

### API Changes

Added:
- Common.iOS.TextFilters
  - ITextFilter
  - GroupFilter
  - LengthFilter
  - ForbiddenCharsFilter
- Common.iOS.Extensions
  - UITextViewExtensions

Removed:
- LengthValidator
- TextFieldWithoutPaste
- UITextFieldExtensions

### Platforms Affected

- iOS

### PR Checklist


Softeq.XToolkit.Common.iOS/TextFilters/GroupFilter.cs
```diff
@@ -0,0 +1,30 @@
...
+using Foundation;
+using UIKit;
+
+namespace Softeq.XToolkit.Common.iOS.TextFilters
+{
+    /// <summary>
+    ///     This filter combines multiple filters.
+    ///     Returns <see langword="true" /> only if all filters return <see langword="true" />.
+    /// </summary>
+    public class GroupFilter : ITextFilter
```
SavikPavel, line 15: Another option to do that is to create an additional extension method like `public static void SetFilter(this UITextField textField, params ITextFilter filter)` and apply current logic there. I'm not sure if this is a better idea than the current one, but this will allow us to get rid of this filter. @wcoder , what do you think?

wcoder: In this case, we should add the same method for `UITextView` or other future extensions like `UISearchBar`. The current implementation can be reused for any controls in one extension per view.

SavikPavel (approved):

PR #431, opened 2021-01-15
Imagepicker droid

sleushunou:


Softeq.XToolkit.WhiteLabel.Essentials.Droid/ImagePicker/IImagePickerActivityResultHandler.cs
```diff
@@ -0,0 +1,16 @@
...
+using Android.App;
+using Android.Content;
+using Android.Graphics;
+
+namespace Softeq.XToolkit.WhiteLabel.Essentials.Droid.ImagePicker
+{
+    public interface IImagePickerActivityResultHandler
+    {
+        void SetActivity(Activity activity);
+        Task<Bitmap?> HandleImagePickerActivityResultAsync(int requestCode, Result resultCode, Intent data, Android.Net.Uri fileUri);
```
TimofeyBurak, line 14: Android.Net.Uri type could be simplified with using

Softeq.XToolkit.WhiteLabel.Essentials.Droid/ImagePicker/ImagePickerActivityResultHandler.cs
```diff
@@ -0,0 +1,62 @@
...
+
+namespace Softeq.XToolkit.WhiteLabel.Essentials.Droid.ImagePicker
+{
+    public sealed class ImagePickerActivityResultHandler : IImagePickerActivityResultHandler
+    {
+        private WeakReferenceEx<Activity> _activityRef;
+
+        public void SetActivity(Activity activity)
+        {
+            _activityRef = new WeakReferenceEx<Activity>(activity);
```
TimofeyBurak, line 21: Seems like we need Activity reference here only for context. Can't we use ICurrentActivity as dependency for that?

Softeq.XToolkit.WhiteLabel.Essentials.Droid/ImagePicker/ImagePickerActivityResultHandler.cs
```diff
@@ -0,0 +1,62 @@
...
+
+            if (uri != null)
+            {
+                bitmap = ImagePickerUtils.GetBitmap(context, uri);
+
+                if (Build.VERSION.SdkInt >= BuildVersionCodes.N)
+                {
+                    using (var stream = ImagePickerUtils.GetContentStream(context, uri))
+                    {
+                        bitmap = ImagePickerUtils.FixRotation(bitmap, new ExifInterface(stream)).Result;
```
TimofeyBurak, line 50: Why don't just await the task?

Softeq.XToolkit.WhiteLabel.Essentials.Droid/ImagePicker/ImagePickerActivityResultHandler.cs
```diff
@@ -0,0 +1,62 @@
...
+                    }
+                }
+                else
+                {
+                    bitmap = ImagePickerUtils.FixRotation(bitmap, new ExifInterface(uri.ToString())).Result;
```
TimofeyBurak, line 55: Why don't just await the task?

TimofeyBurak (changes requested):

Softeq.XToolkit.WhiteLabel.Essentials.Droid/ImagePicker/ImagePickerActivityResultHandler.cs
wcoder: Currently, we use `IContextProvider->CurrentActivity`

Softeq.XToolkit.WhiteLabel.Essentials.Droid/ImagePicker/IImagePickerActivityResultHandler.cs
```diff
@@ -0,0 +1,16 @@
...
+// http://www.softeq.com
+
+using System.Threading.Tasks;
+using Android.App;
+using Android.Content;
+using Android.Graphics;
+
+namespace Softeq.XToolkit.WhiteLabel.Essentials.Droid.ImagePicker
+{
+    public interface IImagePickerActivityResultHandler
```
TimofeyBurak, line 11: Looking on how this interface is going to be used, here is the idea of how it might be enhanced. We need to handle two types of requests + we might need a way to extend logic for new types of requests, so we could implement three separate methods here:
- HandleImagePickerCameraResultAsync(Result resultCode, Intent data, Uri fileUri);
- HandleImagePickerGalleryResultAsync(Result resultCode, Intent data);
- HandleCustomResultAsync(int requestCode, Result resultCode, Intent data);

TimofeyBurak (approved):

PR #437, opened 2021-03-26
Implement ViewModelComponent

TimofeyBurak:
### Description

Implemented a component class, which incapsulate logic of storing/restoring/cleaning ViewModel in/from internal ViewModels storage

### API Changes

Added:
 - class ViewModelComponent<TViewModel> where TViewModel : class

Changed:
 - ViewModelCache can now store view models of any type, not only IViewModelBase

### Platforms Affected

- Android

### PR Checklist


Softeq.XToolkit.WhiteLabel.Droid/ViewComponents/ViewModelComponent.cs
```diff
@@ -0,0 +1,45 @@
...
+    public sealed class ViewModelComponent<TViewModel> where TViewModel : class
+    {
+        private TViewModel? _viewModel;
+
+        public ViewModelComponent(string viewModelKey)
+        {
+            ViewModelKey = viewModelKey;
+        }
+
+        public TViewModel ViewModel => _viewModel ?? throw new InvalidOperationException("ViewModel instance is not assigned");
```
AlexandraMash, line 18: Should we really throw an exception here? Seems that we expect that _viewModel can become null when app is recreated

AlexandraMash (changes requested):

Softeq.XToolkit.WhiteLabel.Droid/ViewComponents/ViewModelComponent.cs
TimofeyBurak: Yes, we expect that at some point _viewModel might be null. But at the same time, we expect that ViewModel property will be used only after _viewModel restored. If we don't throw exception here when _viewModel is null - we will get NRE somewhere later either way

AlexandraMash: And what if I want to check the property to know if I need to restore it?

TimofeyBurak: This seems like an internal logic of restoration - I could add a check inside RestoreViewModel method.

AlexandraMash: Please add a comment to this property mentioning the exception

Softeq.XToolkit.WhiteLabel.Droid/Internal/IViewModelStore.cs
```diff
@@ -1,47 +1,46 @@
...
-        void Clear();
-    }
-}
+
+namespace Softeq.XToolkit.WhiteLabel.Droid.Internal
+{
+    /// <summary>
+    ///     Defines the contract for a store that keeps view models.
+    /// </summary>
+    internal interface IViewModelStore
```
wcoder, line 11: now we don't depend on the ViewModel specific, can be named:
- `IViewModelStore` -> `IStore` or something else
- `TViewModel` -> `T`

Softeq.XToolkit.WhiteLabel.Droid/ViewComponents/ViewModelComponent.cs
```diff
@@ -0,0 +1,45 @@
+﻿// Developed by Softeq Development Corporation
+// http://www.softeq.com
+
+using System;
+using AndroidX.Fragment.App;
+
+namespace Softeq.XToolkit.WhiteLabel.Droid.ViewComponents
+{
+    public sealed class ViewModelComponent<TViewModel> where TViewModel : class
```
wcoder, line 9: Could you please add some info on how to use this component?

Softeq.XToolkit.WhiteLabel.Droid/ViewComponents/ViewModelComponent.cs
wcoder: @TimofeyBurak Could you please mention how to assign this property? (call initialize method)

AlexandraMash (dismissed):

Softeq.XToolkit.WhiteLabel.Droid/ViewComponents/ViewModelComponent.cs
```diff
@@ -0,0 +1,92 @@
...
+        public string ViewModelKey { get; }
+
+        /// <summary>
+        ///     Initializes <see cref="ViewModelComponent{TViewModel}"/> with an instance of ViewModel.
+        /// </summary>
+        /// <param name="viewModel">
+        ///     ViewModel instance to store.
+        /// </param>
+        public void Initialize(TViewModel viewModel)
+        {
```
wcoder, line 52: What about check when already initialized?

And case when:
- new
- .RestoreViewModel
- .Initialize

Softeq.XToolkit.WhiteLabel.Droid/ViewComponents/ViewModelComponent.cs
```diff
@@ -0,0 +1,92 @@
...
+        }
+
+        /// <summary>
+        ///     Restores ViewModel instance from storage, associated with provided FragmentManager.
+        /// </summary>
+        /// <param name="fragmentManager">
+        ///     Fragment manager where ViewModel has been stored.
+        /// </param>
+        public void RestoreViewModel(FragmentManager fragmentManager)
+        {
```
wcoder, line 75: Can we use `fragmentManager` from constructor? and rename methods, like `Save/Restore/Clear`

Softeq.XToolkit.WhiteLabel.Droid/ViewComponents/ViewModelComponent.cs
```diff
@@ -0,0 +1,92 @@
...
+
+namespace Softeq.XToolkit.WhiteLabel.Droid.ViewComponents
+{
+    /// <summary>
+    ///     Component, which encapsulates logic of storing/restoring ViewModel instance within FragmentManager
+    /// </summary>
+    /// <typeparam name="TViewModel">
+    ///     Type of the ViewModel instance
+    /// </typeparam>
+    public sealed class ViewModelComponent<TViewModel> where TViewModel : class
```
wcoder, line 15: It can be better to rename this class, like `ViewModelStoreComponent` or something else

Softeq.XToolkit.WhiteLabel.Droid/ViewComponents/ViewModelComponent.cs
TimofeyBurak: No, we can't. Restoration happens if parent Activity and Fragments are recreated by the system - in this case FragmentManager is also recreated, and old reference become invalid

Softeq.XToolkit.WhiteLabel.Droid/ViewComponents/ViewModelComponent.cs
TimofeyBurak: > What about check when already initialized?
What is expected behaviour in this case in your opinion? I can't say for sure that it should be forbidden, but I can throw an exception in this case

TimofeyBurak: > And case when:
- new
- .RestoreViewModel
- .Initialize

This is not how this class is intended to be used. I'm not sure it can be handled within this class

wcoder: InvalidOperationException can be better than nothing in this case

Softeq.XToolkit.WhiteLabel.Droid/ViewComponents/ViewModelComponent.cs
TimofeyBurak: The idea behind this class is not only to store ViewModel, but completely control ViewModel's lifecycle within Fragment. The main purpose is to provide ViewModel to a Fragment whenever possible, Save/Restore/Clear methods are just callbacks to the Fragment's lifecycle, basically.

Softeq.XToolkit.WhiteLabel.Droid/ViewComponents/ViewModelComponent.cs
TimofeyBurak: Documentation comments were added to the class

Softeq.XToolkit.WhiteLabel.Droid/Internal/IViewModelStore.cs
TimofeyBurak: done

Softeq.XToolkit.WhiteLabel.Droid/ViewComponents/ViewModelComponent.cs
TimofeyBurak: Documentation comments were added to the class

Softeq.XToolkit.WhiteLabel.Droid/ViewComponents/ViewModelComponent.cs
wcoder: `ViewModelComponent` is too general
what about `ViewModelLifecycleComponent`/`ViewModelFragmentLifecycleComponent`?

TimofeyBurak: Again, controlling lifecycle is just a way to achieve main goal, and the main goal of the component - is to safely provide ViewModel instance whenever possible. It is basically a wrapper around ViewModel property in Fragments. It is what Fragments (well, at least DialogFragments) should use instead of storing reference to ViewModel directly. So it is a component, which provides ViewModel instance to whoever uses it.

Softeq.XToolkit.WhiteLabel.Droid/ViewComponents/ViewModelComponent.cs
TimofeyBurak: done

wcoder (approved):

PR #439, opened 2021-05-15
Get rid of explicit interface implementation in StoryboardFrameNavigationService

AlexandraMash:
### Description

For some reason we had explicit implementation of `IFrameNavigationService` in `StoryboardFrameNavigationService`. It wouldn't allow us to override those methods in a descendant but instead it would allow us to reimplement that interface like this:
`class NewStoryboardFrameNavigationService : StoryboardFrameNavigationService, IFrameNavigationService
{
        public void NavigateToFirstPage()
        {
            ...
        }
}`
This creates a separate method and feels very wrong.

This PR reworks implementation of `IFrameNavigationService` in `StoryboardFrameNavigationService` to avoid such possibilities and provide a more correct way to override required methods if needed in future. Additionally `NavigateToFirstPage` methods becomes virtual for that sake.

### API Changes
 
 None

### Platforms Affected

- iOS

### Behavioral/Visual Changes

None

### Before/After Screenshots

Not applicable

### PR Checklist

- [ ] I have read the [CONTRIBUTING](https://github.com/Softeq/XToolkit.WhiteLabel/blob/master/.github/CONTRIBUTING.md) document

### DIFF
--- a/Softeq.XToolkit.WhiteLabel.iOS/Navigation/StoryboardFrameNavigationService.cs
+++ b/Softeq.XToolkit.WhiteLabel.iOS/Navigation/StoryboardFrameNavigationService.cs
@@ -26,9 +26,7 @@ namespace Softeq.XToolkit.WhiteLabel.iOS.Navigation
 
         public bool IsEmptyBackStack => !NavigationController!.ViewControllers.Any();
 
-        bool IFrameNavigationService.IsInitialized => NavigationController != null;
-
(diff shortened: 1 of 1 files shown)


wcoder (approved):

AlexandraMash:
Discussed with @TimofeyBurak that this should be reworked even better because the service implements 2 different interfaces with the same methods which might not be correct. An issue will be created to rework it later.

PR #451, opened 2021-09-13
add GetDialogDismissTask method

sleushunou:


Softeq.XToolkit.WhiteLabel.Droid/Providers/ContextProvider.cs
```diff
@@ -1,22 +1,48 @@
...
         public Context AppContext => Xamarin.Essentials.Platform.AppContext;
+
+        private void OnPlatformActivityStateChanged(object sender, Xamarin.Essentials.ActivityStateChangedEventArgs e)
+        {
+            if (_activityRef.Target != CurrentActivity)
+            {
+                CurrentActivityChanged?.Invoke(this, EventArgs.Empty);
+            }
+
+            _activityRef = WeakReferenceEx.Create(CurrentActivity);
```
wcoder, line 45: what about creating ref in one place, here?
and skip cases when _activityRef = null or CurrentActivity = null

Softeq.XToolkit.WhiteLabel.Droid/Providers/ContextProvider.cs
```diff
@@ -1,22 +1,48 @@
...
     }
 
     public class EssentialsContextProvider : IContextProvider
     {
+        private WeakReferenceEx<Activity> _activityRef;
+
+        public EssentialsContextProvider()
+        {
+            _activityRef = WeakReferenceEx.Create(CurrentActivity);
+            Xamarin.Essentials.Platform.ActivityStateChanged += OnPlatformActivityStateChanged;
```
TimofeyBurak, line 29: I'd say we could also expose original `ActivityStateChanged` event here, and let the user of the API decide how to use it.. 
BTW, this way we could implement method `WaitForCurrentActivityChangeAsync()` as an extension to IContextProvider interface

Softeq.XToolkit.WhiteLabel.Droid/Providers/ContextProvider.cs
```diff
@@ -1,22 +1,49 @@
...
 
+    public class ActivityStateChangedEventArgs : EventArgs
+    {
+        internal ActivityStateChangedEventArgs(Activity activity, ActivityState state)
+        {
+            State = state;
+            Activity = activity;
+        }
+
+        public ActivityState State { get; }
```
wcoder, line 30: `ActivityState` is enum from Essentials

wcoder: we can use `Lifecycle.State` from AndroidX

TimofeyBurak (approved):

PR #454, opened 2021-11-08
Feature/container collection resolve

nkrisko:
### Description

Added possibility to register 'Singleton Type' on Bootstrapper for support to store single type with different interfaces

### API Changes

Added:
 - void DryIocContainerAdapter.SingletonType\<TType>(IfRegistered ifRegistered = IfRegistered.AppendNew);
 - void DryIocContainerBuilder.SingletonType\<TType>(IfRegistered ifRegistered = IfRegistered.AppendNew);

### Platforms Affected

- Core (all platforms)
- iOS
- Android

### Behavioral/Visual Changes

None

### Before/After Screenshots

Not applicable

### PR Checklist


wcoder (commented):
As I understand DryIoc already provide this behavior via wrappers:

```cs
class ConsumerClass {
    ConsumerClass(IList<IService> services) {}
}
```
Details: https://github.com/dadhi/DryIoc/blob/master/docs/DryIoc.Docs/Wrappers.md#ienumerable-or-array-of-a

nkrisko:
@wcoder You are right, commit partly has been reverted to exist version. Only methods for registration interfaces for instance type was stayed.

Softeq.XToolkit.WhiteLabel/Bootstrapper/Abstract/IContainerBuilder.cs
```diff
@@ -29,6 +29,8 @@ public interface IContainerBuilder
 
         void Singleton<TService>(Func<IContainer, TService> func, IfRegistered ifRegistered = IfRegistered.AppendNew);
 
+        void SingletonType<TType>(IfRegistered ifRegistered = IfRegistered.AppendNew);
```
wcoder, line 32: The method name doesn't describe it enough to use and override (could you please rename and add documentation?)

wcoder (changes requested):
Are we really need this API? Or it can be realized via:
- split the class into the separate implementations and register them
- create one instance and register it for all interfaces
- create and register extended IContainerBuilder implementation for the concrete case/project

Cons:
- only DryIoc has such a simple API, other containers describe this behavior as more complex with more additional types

Softeq.XToolkit.WhiteLabel/Bootstrapper/Abstract/IContainerBuilder.cs
TimofeyBurak: Autofac also has extension to register instance as all of implemented interfaces: https://autofac.org/apidoc/html/5BE7EA96.htm Which other DI implementation do we need to consider to be potentially supported?
Also, regarding the custom `IContainerBuilder` interface, - in this case we would have to also implement brand new `Bootstrapper` hierarchy, or extend `Bootstrapper` to support any container implementation

TimofeyBurak:
> Are we really need this API? Or it can be realized via:
> 
> * split the class into the separate implementations and register them
> * create one instance and register it for all interfaces
> * create and register extended IContainerBuilder implementation for the concrete case/project
> 
> Cons:
> 
> * only DryIoc has such a simple API, other containers describe this behavior as more complex with more additional types

Autofac also has extension to register instance as all of implemented interfaces: https://autofac.org/apidoc/html/5BE7EA96.htm Which other DI implementation do we need to consider to be potentially supported?
Also, regarding the custom `IContainerBuilder` interface, - in this case we would have to also implement brand new `Bootstrapper` hierarchy, or extend `Bootstrapper` to support any container implementation

wcoder:
@TimofeyBurak 
> Which other DI implementation do we need to consider to be potentially supported?

For WL containers I'm considering the following:
- DryIoc (as best for now)
- AutoFac (as an alternative for migrating old projects)
- Microsoft DI and MS.DI.Abstractions (as next generation)
- any other container (simple interface - easy integration)

> Autofac also has extension to register instance as all of implemented interfaces: https://autofac.org/apidoc/html/5BE7EA96.htm 

Cool, I didn't know about this method.

> in this case we would have to also implement brand new Bootstrapper hierarchy, or extend Bootstrapper to support any container implementation

Yes, that should work.

PR #479, opened 2022-04-13
Fix closing ios dialogs

dimysa:
### Description

It is possible that the controller that presented the dialog has been removed from the navigation stack, so calling `DismissViewController` on it will not close the previously shown dialog. ~~We can call `DismissViewController` on the shown `ViewController` and `UIKit` will do all the work for us~~ (as it turned out, not always - in the case if the dismiss controller, that presented another one, then it will not dismiss itself, but the presented one)
https://developer.apple.com/documentation/uikit/uiviewcontroller/1621505-dismiss

### API Changes

 
 None

### Platforms Affected

- iOS

### Behavioral/Visual Changes

None

### Before/After Screenshots

Before
Simulator Screen Recording - iPhone 8 - 2022-04-13 at 19 00 29

After
Simulator Screen Recording - iPhone 8 - 2022-04-13 at 19 01 10

### PR Checklist

### DIFF
--- a/Softeq.XToolkit.WhiteLabel.iOS/Services/StoryboardDialogsService.cs
+++ b/Softeq.XToolkit.WhiteLabel.iOS/Services/StoryboardDialogsService.cs
@@ -172,7 +172,8 @@ namespace Softeq.XToolkit.WhiteLabel.iOS.Services
                 try
                 {
                     presentedController.View?.EndEditing(true);
-                    presentedController.DismissViewController(true, () => tcs.TrySetResult(true));
+                    presentedController.PresentingViewController
+                        .DismissViewController(true, () => tcs.TrySetResult(true));
                 }
                 catch (Exception ex)
                 {
--- a/Softeq.XToolkit.WhiteLabel.iOS/Services/StoryboardDialogsService.cs
+++ b/Softeq.XToolkit.WhiteLabel.iOS/Services/StoryboardDialogsService.cs
@@ -154,11 +154,18 @@ namespace Softeq.XToolkit.WhiteLabel.iOS.Services
 
             Execute.BeginOnUIThread(() =>
             {
-                var targetViewController = _viewLocator.GetView(viewModel);
-                var topViewController = _viewLocator.GetTopViewController();
-                topViewController.View?.EndEditing(true);
-                topViewController.PresentViewController(targetViewController, true, null);
-                tcs.TrySetResult(targetViewController);
+                try
+                {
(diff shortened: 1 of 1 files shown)


wcoder (commented):
Could you please check the case from the previous fix of this part #208 ?

dimysa:
> Could you please check the case from the previous fix of this part #208 ?

Yes, it looks like we need another solution

dimysa:
@wcoder Found another solution, maybe a little weird, but it works for my case and for #208

wcoder (commented):
@dimysa What iOS version is used for you?
I tested the `master` branch for iOS 13.5, 15.4 - everything works fine.

wcoder:
> master branch iOS 14.0 and 15.4

Hm, how can I reproduce **Before** state?

dimysa:
Like this
```
private void OpenTwoDialogs()
{
    _dialogsService
        .For<SimpleDialogPageViewModel>()
        .WithParam(x => x.FirstName, "First1")
        .WithParam(x => x.LastName, "Last1")
        .Navigate<Person?>()
    .FireAndForget();
    _pageNavigationService.GoBack();
}
```

Softeq.XToolkit.WhiteLabel.iOS/Services/StoryboardDialogsService.cs
```diff
@@ -148,31 +148,32 @@ public virtual Task<string> ShowDialogAsync(ActionSheetDialogConfig config)
...
+            var tcs = new TaskCompletionSource<UIViewController>();
 
             Execute.BeginOnUIThread(() =>
             {
                 var targetViewController = _viewLocator.GetView(viewModel);
                 var topViewController = _viewLocator.GetTopViewController();
                 topViewController.View?.EndEditing(true);
                 topViewController.PresentViewController(targetViewController, true, null);
-                tcs.TrySetResult(new PresentedViewController(topViewController, targetViewController));
+                tcs.TrySetResult(targetViewController);
```
wcoder, line 161: Suggestion:
```cs
try
{
    var targetViewController = _viewLocator.GetView(viewModel);
    var topViewController = _viewLocator.GetTopViewController();
    topViewController.View?.EndEditing(true);
    topViewController.PresentViewController(targetViewController, true, () => tcs.TrySetResult(targetViewController));
}
catch (Exception ex)
{
    _logger.Error(ex);
    tcs.TrySetException(ex);
}
```

wcoder (dismissed):

TimofeyBurak (approved):

PR #504, opened 2022-11-22
add token synchronizer

sleushunou:
### Issues Resolved

- fixes #

### API Changes

 None

### Platforms Affected

- Core (all platforms)
- iOS
- Android

### Behavioral/Visual Changes

None

### Before/After Screenshots

Not applicable

### PR Checklist

### DIFF
--- a/Softeq.XToolkit.PushNotifications.Droid/Services/DroidPushNotificationsConsumer.cs
+++ b/Softeq.XToolkit.PushNotifications.Droid/Services/DroidPushNotificationsConsumer.cs
@@ -5,7 +5,6 @@ using System.Threading.Tasks;
 using Android.Content;
 using AndroidX.Lifecycle;
 using Firebase.Messaging;
-using Softeq.XToolkit.Common.Extensions;
 using Softeq.XToolkit.Common.Logger;
 using Softeq.XToolkit.PushNotifications.Abstract;
 using Softeq.XToolkit.PushNotifications.Droid.Abstract;
@@ -17,10 +16,8 @@ namespace Softeq.XToolkit.PushNotifications.Droid.Services
     /// </summary>
     public sealed class DroidPushNotificationsConsumer : IDroidPushNotificationsConsumer
     {
-        private readonly IPushTokenStorageService _pushTokenStorageService;
         private readonly IDroidPushNotificationsParser _pushNotificationsParser;
         private readonly IPushNotificationsHandler _pushNotificationsHandler;
-        private readonly IRemotePushNotificationsService _remotePushNotificationsService;
         private readonly IDroidPushNotificationPresenter _pushNotificationPresenter;
         private readonly IPushTokenSynchronizer _pushTokenSynchronizer;
         private readonly ILogger _logger;
@@ -31,24 +28,19 @@ namespace Softeq.XToolkit.PushNotifications.Droid.Services
         ///     Initializes a new instance of the <see cref="DroidPushNotificationsConsumer"/> class.
         /// </summary>
         /// <param name="pushNotificationsParser">Parses remote messages into common format.</param>
-        /// <param name="pushTokenStorageService">Stores push notification token and it's status.</param>
         /// <param name="pushNotificationsHandler">Handles received push notifications events.</param>
-        /// <param name="remotePushNotificationsService">Propagates push notification token to remote server.</param>
         /// <param name="pushNotificationPresenter">Responsible for presenting push notification.</param>
+        /// <param name="pushTokenSynchronizer">Responsible for push token lifecycle and server syncronization.</param>
         /// <param name="logManager">Provides logging.</param>
         public DroidPushNotificationsConsumer(
             IDroidPushNotificationsParser pushNotificationsParser,
-            IPushTokenStorageService pushTokenStorageService,
             IPushNotificationsHandler pushNotificationsHandler,
-            IRemotePushNotificationsService remotePushNotificationsService,
             IDroidPushNotificationPresenter pushNotificationPresenter,
             IPushTokenSynchronizer pushTokenSynchronizer,
             ILogManager logManager)
         {
             _pushNotificationsParser = pushNotificationsParser;
-            _pushTokenStorageService = pushTokenStorageService;
             _pushNotificationsHandler = pushNotificationsHandler;
-            _remotePushNotificationsService = remotePushNotificationsService;
             _pushNotificationPresenter = pushNotificationPresenter;
             _pushTokenSynchronizer = pushTokenSynchronizer;
             _logger = logManager.GetLogger<DroidPushNotificationsConsumer>();
@@ -88,19 +80,13 @@ namespace Softeq.XToolkit.PushNotifications.Droid.Services
         /// <inheritdoc />
         public void OnPushTokenRefreshed(string token)
         {
-            var currentToken = _pushTokenStorageService.PushToken;
-            var hasNewToken = !string.IsNullOrEmpty(currentToken) && currentToken != token;
-
-            if (hasNewToken)
-            {
-                OnRegisteredForPushNotifications(token);
-            }
+            _pushTokenSynchronizer.OnTokenChanged(token);
         }
 
         /// <inheritdoc />
         public Task OnUnregisterFromPushNotifications()
         {
-            return UnregisterFromRemotePushNotifications();
+            return _pushTokenSynchronizer.UnregisterFromRemotePushNotificationsAsync();
         }
 
         private void OnMessageReceivedInternal(PushNotificationModel parsedNotification, bool inForeground)
@@ -112,55 +98,6 @@ namespace Softeq.XToolkit.PushNotifications.Droid.Services
             else
             {
                 _pushNotificationsHandler.HandlePushNotificationReceived(parsedNotification, inForeground);
-            }
-        }
-
-        private void OnRegisteredForPushNotifications(string token)
-        {
-            if (string.IsNullOrWhiteSpace(token))
-            {
-                OnRegisterFailedInternal().FireAndForget(_logger);
-            }
-            else
-            {
-                OnRegisterSuccessInternal(token).FireAndForget(_logger);
-            }
-        }
-
-        private async Task OnRegisterSuccessInternal(string token)
-        {
-            await _pushTokenSynchronizer.OnRegisterSuccessInternalAsync(token)
-                .ConfigureAwait(false);
-
-            _pushNotificationsHandler.OnPushRegistrationCompleted(
-                _pushTokenStorageService.IsTokenRegisteredInSystem,
-                _pushTokenStorageService.IsTokenSavedOnServer);
-        }
-
-        private async Task OnRegisterFailedInternal()
-        {
-            _pushTokenSynchronizer.OnRegisterFailedInternal();
-
-            await UnregisterFromRemotePushNotifications();
-
-            _pushNotificationsHandler.OnPushRegistrationCompleted(
-                _pushTokenStorageService.IsTokenRegisteredInSystem,
-                _pushTokenStorageService.IsTokenSavedOnServer);
-        }
-
-        private async Task UnregisterFromRemotePushNotifications()
-        {
-            if (!_pushTokenStorageService.IsTokenSavedOnServer)
-            {
-                return;
-            }
-
-            var tokenRemovedFromServer = await _remotePushNotificationsService
-                .RemovePushNotificationsToken(_pushTokenStorageService.PushToken)
-                .ConfigureAwait(false);
-            if (tokenRemovedFromServer)
-            {
-                _pushTokenStorageService.IsTokenSavedOnServer = false;
             }
         }
     }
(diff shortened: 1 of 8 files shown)


Softeq.XToolkit.PushNotifications.Droid/Services/DroidPushNotificationsConsumer.cs
```diff
@@ -41,13 +42,15 @@ public sealed class DroidPushNotificationsConsumer : IDroidPushNotificationsCons
             IPushNotificationsHandler pushNotificationsHandler,
             IRemotePushNotificationsService remotePushNotificationsService,
             IDroidPushNotificationPresenter pushNotificationPresenter,
+            IPushTokenSynchronizer pushTokenSynchronizer,
```
TimofeyBurak, line 45: Can you please also update constructor documentation, add new parameter

Softeq.XToolkit.PushNotifications.iOS/Services/IosPushNotificationsConsumer.cs
```diff
@@ -38,6 +39,7 @@ public sealed class IosPushNotificationsConsumer : IIosPushNotificationsConsumer
         /// <param name="showForegroundNotificationsInSystemOptions">Defines how push notification should be presented.</param>
         /// <param name="logManager">Provides logging.</param>
         public IosPushNotificationsConsumer(
+            IPushTokenSynchronizer pushTokenSynchronizer,
```
TimofeyBurak, line 42: Can you please also update constructor documentation, add new parameter

Softeq.XToolkit.PushNotifications/PushTokenSynchronizer.cs
```diff
@@ -0,0 +1,76 @@
+﻿// Developed for PAWS-HALO by Softeq Development Corporation
+// http://www.softeq.com
+
+using System;
+using System.Threading.Tasks;
+using Softeq.XToolkit.PushNotifications.Abstract;
+
+namespace Softeq.XToolkit.PushNotifications
+{
+    public class PushTokenSynchronizer : IPushTokenSynchronizer
```
TimofeyBurak, line 10: I would be better, if `PushTokenSynchronizer` would handle everything connected to communication with backend, including "unregistration"

TimofeyBurak (changes requested):

TimofeyBurak:
A few notes regarding the idea of "synchroniser":
- it is intended to be an abstraction layer between system and backend, but unlike `IRemotePushNotificationsService` it controls not only the means of transportation token to backend, but the whole lifecycle of the token from when it is registered inn system, until it is "unregistered";
- as an abstraction level, it should be the only entry point for all the work connected to the "synchronisation" of the token. It means that consumers should communicate with synchroniser, but not with `IRemotePushNotificationsService` and `IPushTokenStorageService`. Otherwise abstraction won't make any sense, as it would be impossible to implement this abstraction without using `IRemotePushNotificationsService` and `IPushTokenStorageService`. But this would mean that some of the logic might be moved to the synchroniser too, for instance:

https://github.com/Softeq/XToolkit.WhiteLabel/blob/4128fb593159fffcb1966f6b69ebd78785d6bcd6/Softeq.XToolkit.PushNotifications.Droid/Services/DroidPushNotificationsConsumer.cs#L91-L92
here consumers checks whether token is new or not - but it should be synchroniser who decide if token need to be sent to backend;

https://github.com/Softeq/XToolkit.WhiteLabel/blob/4128fb593159fffcb1966f6b69ebd78785d6bcd6/Softeq.XToolkit.PushNotifications.Droid/Services/DroidPushNotificationsConsumer.cs#L135-L137
here `IPushTokenStorageService` says whether token was saved in system/on backend - but again, it should be synchroniser who responsible for that and who knows that for sure (the value can be returned from `IPushTokenSynchronizer.OnRegisterSuccessInternalAsync()`, for instance). At least the backend part is responsibility of synchroniser, registration status in system is known by the consumer itself, not sure why we store it;

https://github.com/Softeq/XToolkit.WhiteLabel/blob/4128fb593159fffcb1966f6b69ebd78785d6bcd6/Softeq.XToolkit.PushNotifications.Droid/Services/DroidPushNotificationsConsumer.cs#L151-L165
here consumer "unregisters" token on backend, but again it should be a responsibility of the synchroniser

wcoder (changes requested):
Could you describe the problem this "synchronizer" concept tries to solve?

TimofeyBurak:
@wcoder original implementation does not allowing any means of retrying to send token to backend. Also, there was no way of restarting the process of synchronisation after relaunch (if token is not yet synchronised)

Softeq.XToolkit.PushNotifications/PushTokenSynchronizer.cs
```diff
@@ -0,0 +1,153 @@
...
+            if (IsTokenRegisteredInSystem && !IsTokenSavedOnServer)
+            {
+                return DoSendTokenToServer(_pushTokenStorageService.PushToken);
+            }
+
+            return Task.CompletedTask;
+        }
+
+        /// <inheritdoc />
+        public async Task UnregisterFromRemotePushNotificationsAsync()
```
TimofeyBurak, line 72: Seems like we need to change `IsRegisteredInSystem` in this case

Softeq.XToolkit.PushNotifications/PushTokenSynchronizer.cs
```diff
@@ -0,0 +1,153 @@
...
+                true,
+                IsTokenSavedOnServer);
+        }
+
+        private async Task DoSendTokenToServer(string token)
+        {
+            IsTokenSavedOnServer = await _remotePushNotificationsService
+                .SendPushNotificationsToken(token).ConfigureAwait(false);
+
+            int attemptsCount = 1;
```
TimofeyBurak, line 141: `var` can be used

Softeq.XToolkit.PushNotifications/PushTokenSynchronizer.cs
```diff
@@ -0,0 +1,153 @@
...
+        }
+
+        private bool IsTokenRegisteredInSystem
+        {
+            get => _internalSettings.GetValueOrDefault(_isTokenRegisteredInSystemKey, default(bool));
+            set => _internalSettings.AddOrUpdateValue(_isTokenRegisteredInSystemKey, value);
+        }
+
+        /// <inheritdoc />
+        public Task ResendTokenToServerIfNeedAsync()
```
TimofeyBurak, line 61: One more thing we could check here is if token is unregistered in system, but still saved on backend. In this case method should be called something like `SynchronizeTokenIfNeededAsync()`

TimofeyBurak (changes requested):

Softeq.XToolkit.PushNotifications/PushTokenSynchronizer.cs
```diff
@@ -0,0 +1,76 @@
+﻿// Developed for PAWS-HALO by Softeq Development Corporation
```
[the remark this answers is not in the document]
sleushunou: Fixed

Softeq.XToolkit.PushNotifications.iOS/Services/IosPushNotificationsConsumer.cs
sleushunou: Fixed

Softeq.XToolkit.PushNotifications/PushTokenSynchronizer.cs
sleushunou: Fixed

Softeq.XToolkit.PushNotifications/PushTokenSynchronizer.cs
sleushunou: Fixed

Softeq.XToolkit.PushNotifications/PushTokenSynchronizer.cs
sleushunou: Fixed

Softeq.XToolkit.PushNotifications/PushTokenSynchronizer.cs
sleushunou: Fixed

Softeq.XToolkit.PushNotifications/Abstract/IPushTokenSynchronizer.cs
```diff
@@ -0,0 +1,16 @@
+﻿// Developed for PAWS-HALO by Softeq Development Corporation
```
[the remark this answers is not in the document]
TimofeyBurak: Shouldn't we use
```
// Developed by Softeq Development Corporation
// http://www.softeq.com
```
as a header?

Softeq.XToolkit.PushNotifications/PushTokenSynchronizer.cs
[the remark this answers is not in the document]
TimofeyBurak: Shouldn't we use
```
// Developed by Softeq Development Corporation
// http://www.softeq.com
```
as a header?

TimofeyBurak (approved):

Softeq.XToolkit.PushNotifications/Abstract/IPushTokenSynchronizer.cs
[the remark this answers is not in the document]
sleushunou: Fixed

Softeq.XToolkit.PushNotifications/PushTokenSynchronizer.cs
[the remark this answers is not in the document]
sleushunou: Fixed

Softeq.XToolkit.PushNotifications/PushTokenSynchronizer.cs
```diff
@@ -0,0 +1,161 @@
...
+            IsTokenSavedOnServer = await _remotePushNotificationsService
+                .SendPushNotificationsToken(token).ConfigureAwait(false);
+
+            var attemptsCount = 1;
+
+            while (!IsTokenSavedOnServer && attemptsCount < MaxAttemptsCount)
+            {
+                await Task.Delay(TokenSendRetryDelay)
+                    .ConfigureAwait(false);
+                IsTokenSavedOnServer = await _remotePushNotificationsService
```
wcoder, line 155: can we guarantee valid states in the case when sync was called twice in parallel?
```cs
SynchronizeTokenIfNeededAsync();
SynchronizeTokenIfNeededAsync();
```
`IsTokenSavedOnServer` is unpredictably in this case

Softeq.XToolkit.PushNotifications/PushTokenSynchronizer.cs
```diff
@@ -0,0 +1,161 @@
...
+                await DoSendTokenToServer(token)
+                    .ConfigureAwait(false);
+            }
+
+            _pushNotificationsHandler.OnPushRegistrationCompleted(
+                true,
+                IsTokenSavedOnServer);
+        }
+
+        private async Task DoSendTokenToServer(string token)
```
wcoder, line 144: Why can't we move this retry logic to `IRemotePushNotificationsService` implementation?

Softeq.XToolkit.PushNotifications/PushTokenSynchronizer.cs
```diff
@@ -0,0 +1,161 @@
...
+using System;
+using System.Threading.Tasks;
+using Softeq.XToolkit.Common.Extensions;
+using Softeq.XToolkit.Common.Interfaces;
+using Softeq.XToolkit.Common.Logger;
+using Softeq.XToolkit.PushNotifications.Abstract;
+
+namespace Softeq.XToolkit.PushNotifications
+{
+    public class PushTokenSynchronizer : IPushTokenSynchronizer
```
wcoder, line 13: Looks like this is `IPushTokenManager` (token delegate + token synchronizer)
@TimofeyBurak what do you think?

TimofeyBurak: You mean naming-wise? I'm not sure, "manager" to me sounds like a class that manages something, meaning controls something. This class does not control "PushToken", it only reacts to the changes made by other classes. If this class is controlling something - it's "PushToken Synchronisation", thus `IPushTokenSyncronizationManager` could be possible, but `IPushTokenSynchronizer` seems shorter with the same meaning)

Softeq.XToolkit.PushNotifications/PushTokenSynchronizer.cs
TimofeyBurak: Retry logic can be implemented on both sides, I guess, but `IRemotePushNotificationsService` doesn't know about token lifecycle, so it won't be abele to stop retrying to send token if we already "unregistered" this token.
`IPushTokenSynchronizer` does know about token lifecycle, so it can stop retrying on "unregister" (although it is not implemented yet)

Softeq.XToolkit.PushNotifications/PushTokenSynchronizer.cs
```diff
@@ -0,0 +1,161 @@
...
+            }
+
+            _pushNotificationsHandler.OnPushRegistrationCompleted(
+                true,
+                IsTokenSavedOnServer);
+        }
+
+        private async Task DoSendTokenToServer(string token)
+        {
+            IsTokenSavedOnServer = await _remotePushNotificationsService
```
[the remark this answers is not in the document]
sleushunou: Fixed

Softeq.XToolkit.PushNotifications/PushTokenSynchronizer.cs
```diff
@@ -132,7 +139,9 @@ private async Task OnRegisterSuccessInternalAsync(string token)
             {
                 _pushTokenStorageService.PushToken = token;
 
-                await DoSendTokenToServer(token)
+                _doSendToServerCts?.Cancel();
```
wcoder, line 142: safe value setting that can be moved inside a method:
```cs
Interlocked.Exchange(ref _cts, new CancellationTokenSource())?.Cancel();
```

wcoder (approved):

PR #507, opened 2022-12-13
added ability to open 2 or more dialogs in a row

vadimpylsky:


Softeq.XToolkit.WhiteLabel.Droid/Dialogs/DialogFragmentBase.cs
```diff
@@ -122,6 +131,13 @@ public void Show()
             Show(fragmentManager, null);
         }
 
+        public override void OnSaveInstanceState(Bundle outState)
+        {
+            base.OnSaveInstanceState(outState);
+
+            outState.PutString(ViewModelKey, _key);
```
wcoder, line 138: please use `GetKey()`

wcoder (commented):
Looks like all these changes can be made in the child class, we need only change the signature for method `GetKey`. What do you think?

PR #511, opened 2023-01-12
Removed executePendingTransaction call from the navigation service

dimysa:
### Description

Calling `executePendingTransaction` if we want to swap two fragments will very quickly lead to an `IllegalStateException: FragmentManager is already executing transactions`, so it should be removed. It looks like just calling `commit`, although asynchronous, won't cause problems, since each `commit` is queued and executed when the main thread is ready
[docs](https://developer.android.com/reference/androidx/fragment/app/FragmentTransaction#commit())

### API Changes

 None

### Platforms Affected

- Android

### Behavioral/Visual Changes

None

### Before/After Screenshots

Not applicable

### PR Checklist

### DIFF
--- a/Softeq.XToolkit.WhiteLabel.Droid/Navigation/DroidFrameNavigationService.cs
+++ b/Softeq.XToolkit.WhiteLabel.Droid/Navigation/DroidFrameNavigationService.cs
@@ -227,8 +227,6 @@ namespace Softeq.XToolkit.WhiteLabel.Droid.Navigation
                 .BeginTransaction()
                 .Replace(config.ContainerId, fragment);
 
-            var preparedTransaction = PrepareTransaction(transaction);
-
             PrepareTransaction(transaction).Commit();
         }
     }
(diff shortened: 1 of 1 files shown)


Softeq.XToolkit.WhiteLabel.Droid/Navigation/DroidFrameNavigationService.cs
```diff
@@ -229,15 +229,7 @@ private void SetCurrentFragment(FrameNavigationConfig config, IViewModelBase top
...
-            Execute.BeginOnUIThread(() =>
-            {
-                // Commit() method is executed asynchronously, so there's no way to know when will the transaction actually complete.
-                // At the same time, trying to execute multiple transactions at the same time will result in exception.
-                // ExecutePendingTransactions() should prevent exceptions even if two transactions are initiated at the same time
-                // IMPORTANT: ExecutePendingTransactions() is required to me called from UI thread
-                config.Manager.ExecutePendingTransactions();
-                preparedTransaction.Commit();
-            });
+            PrepareTransaction(transaction).Commit();
```
TimofeyBurak, line 232: `PrepareTransaction()` is called twice now

TimofeyBurak (changes requested):

Softeq.XToolkit.WhiteLabel.Droid/Navigation/DroidFrameNavigationService.cs
dimysa: Oops. Fixed

TimofeyBurak (approved):

PR #512, opened 2023-02-14
Fixed handling of silent push notifications on Android

dimysa:
### Description

Now, when we close the Android app and send a silent notification (where `notification = null`) it will not be handled because the `OnNotificationReceived` delegate will be `null`, since no one will create/resolve the `DroidPushNotificationsService`. Since the `FirebaseMessagingService` is recreated if the app is closed, I simply added getting the `DroidPushNotificationsService` from the container since the container is also created since the `Android.App.Application` object is created when the `FirebaseMessagingService` is recreated

### API Changes

 None

### Platforms Affected

- Android

### Behavioral/Visual Changes

None

### Before/After Screenshots

Not applicable

### PR Checklist

### DIFF
--- a/Softeq.XToolkit.PushNotifications.Droid/Abstract/IDroidFirebaseMessagingHandler.cs
+++ b/Softeq.XToolkit.PushNotifications.Droid/Abstract/IDroidFirebaseMessagingHandler.cs
@@ -1,0 +1,25 @@
+﻿// Developed for PAWS-HALO by Softeq Development Corporation
+// http://www.softeq.com
+
+using Firebase.Messaging;
+
+namespace Softeq.XToolkit.PushNotifications.Abstract
+{
+    /// <summary>
+    /// Basic interface for the event handler from the <see cref="FirebaseMessagingService"/>
+    /// </summary>
+    internal interface IDroidFirebaseMessagingHandler
+    {
+        /// <summary>
+        ///    Callback called when push notification is received.
+        /// </summary>
+        /// <param name="message">Received remote message.</param>
+        void OnNotificationReceived(RemoteMessage message);
+
+        /// <summary>
+        ///     Callback called when push notification token has changed.
+        /// </summary>
+        /// <param name="token">New push notification token.</param>
+        void OnPushTokenRefreshed(string token);
+    }
+}
--- a/Softeq.XToolkit.PushNotifications.Droid/Abstract/IDroidFirebaseMessagingHandler.cs
+++ b/Softeq.XToolkit.PushNotifications.Droid/Abstract/IDroidFirebaseMessagingHandler.cs
@@ -1,4 +1,4 @@
-﻿// Developed for PAWS-HALO by Softeq Development Corporation
+﻿// Developed by Softeq Development Corporation
 // http://www.softeq.com
 
 using Firebase.Messaging;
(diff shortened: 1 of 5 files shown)


Softeq.XToolkit.PushNotifications.Droid/XFirebaseMessagingService.cs
```diff
@@ -4,28 +4,35 @@
...
-    public class XFirebaseMessagingService : FirebaseMessagingService
+    internal class XFirebaseMessagingService : FirebaseMessagingService
     {
-        public static Action<string>? OnTokenRefreshed;
-        public static Action<RemoteMessage>? OnNotificationReceived;
+        private readonly IDroidFirebaseMessagingHandler _messagingHandler;
+
+        public XFirebaseMessagingService()
+        {
+            _messagingHandler = Dependencies.Container.Resolve<DroidPushNotificationsService>();
```
wcoder, line 21: looks like should be `IDroidFirebaseMessagingHandler`

wcoder, line 21: are we sure that the container is ready for use at this time?

wcoder, line 21: looks like should be `IDroidFirebaseMessagingHandler`

Softeq.XToolkit.PushNotifications.Droid/Abstract/IDroidFirebaseMessagingHandler.cs
```diff
@@ -0,0 +1,25 @@
+﻿// Developed for PAWS-HALO by Softeq Development Corporation
```
wcoder, line 1: Use xtoolkit header

wcoder (changes requested):

Softeq.XToolkit.PushNotifications.Droid/XFirebaseMessagingService.cs
dimysa: Yes, the `Application` class where the container is initialised will be created before this class is created

Softeq.XToolkit.PushNotifications.Droid/Abstract/IDroidFirebaseMessagingHandler.cs
dimysa: Fixed

Softeq.XToolkit.PushNotifications.Droid/XFirebaseMessagingService.cs
dimysa: Here are my thoughts
`IDroidFirebaseMessagingHandler` doesn't make sense to make it public, so it can't be registered in a container.
Right now the `DroidPushNotificationsService` is (or should be) registered in the container like this
```
builder.Singleton<DroidPushNotificationsService>();
builder.Singleton<IPushNotificationsService>(container => container.Resolve<DroidPushNotificationsService>());
builder.Singleton<IActivityLauncherDelegate>(container => container.Resolve<DroidPushNotificationsService>());
```
So we can get the `DroidPushNotificationsService` from the container and use it internally as a message handler from the `FirebaseMessagingService`, without showing it to the outside

wcoder (approved):

PR #527, opened 2023-06-07
Implemented differents methods for creating and caching HttpMessageHandler

dimysa:
### Description

Now, if we go in to get a new HttpMessageHandler for another HttpClient, we get the same HttpMessageHandler. And if HttpClient is disposed, then HttpMessageHandler will also be disposed, which will make other HttpClients not work

### API Changes

Added:
 - bool HttpMessageHandlerProvider.CreateOrGetCachedDefaultHandler()

Changed:
 - bool HttpMessageHandlerProvider.CreateDefaultHandler()

### Platforms Affected

- Core (all platforms)

### Behavioral/Visual Changes

None

### Before/After Screenshots

Not applicable

### PR Checklist

### DIFF
--- a/Softeq.XToolkit.Remote.Tests/Client/HttpMessageHandlerProviderTests.cs
+++ b/Softeq.XToolkit.Remote.Tests/Client/HttpMessageHandlerProviderTests.cs
@@ -12,9 +12,35 @@ namespace Softeq.XToolkit.Remote.Tests.Client
         [Fact]
         public void CreateDefaultHandler_Always_ReturnsHttpMessageHandler()
         {
+            var result = HttpMessageHandlerProvider.CreateOrGetCachedDefaultHandler();
+
+            Assert.IsAssignableFrom<HttpMessageHandler>(result);
+        }
+
+        [Fact]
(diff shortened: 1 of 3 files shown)


Softeq.XToolkit.Remote.Tests/Client/HttpMessageHandlerProviderTests.cs
```diff
@@ -11,10 +11,36 @@ public class HttpMessageHandlerProviderTests
...
         [Fact]
         public void CreateDefaultHandler_Always_ReturnsHttpMessageHandler()
+        {
+            var result = HttpMessageHandlerProvider.CreateOrGetCachedDefaultHandler();
+
+            Assert.IsAssignableFrom<HttpMessageHandler>(result);
+        }
+
+        [Fact]
+        public void CreateOrGetCachedDefaultHandler_Always_ReturnsHttpMessageHandler()
```
TimofeyBurak, line 21: This test (and the one above) actually don't test anything (or it tests that C# is a strongly typed language 🙂) - we already know that the result is of type `HttpMessageHandler` because it is the return type of the method.
The idea might be to check if result is not null - that might have more cense

TimofeyBurak (approved):

PR #531, opened 2023-08-03
Replace Connectivity plugin to MAUI Essentials

wcoder:
### Description

Replace the deprecated Xam.Plugin.Connectivity plugin to MAUI Essentials.

### API Changes

Changed:
 - `IEnumerable<ConnectionType> IConnectivityService.ConnectionTypes` => `IEnumerable<ConnectionProfile> IConnectivityService.ConnectionProfiles`
 - `ConnectivityService` => `EssentialsConnectivityService`
 - `IosConnectivityService`

 Removed:
 - `event IConnectivityService.ConnectivityTypeChanged`
 - `bool IConnectivityService.IsSupported`

### Platforms Affected

- Core (all platforms)
- iOS
- Android

### Behavioral/Visual Changes

None

### Before/After Screenshots

Not applicable

### PR Checklist


Softeq.XToolkit.Connectivity/IConnectivityService.cs
```diff
@@ -3,20 +3,28 @@
...
 using System.Collections.Generic;
-using Plugin.Connectivity.Abstractions;
+using Microsoft.Maui.Networking;
 
 namespace Softeq.XToolkit.Connectivity
 {
+    /// <summary>
+    ///     Interface for Connectivity Service.
+    /// </summary>
     public interface IConnectivityService : IDisposable
```
TimofeyBurak, line 13: Maybe we should remove `Dispose()` method from the service interface?
- implementations could still implement Disposable pattern as they do, and IoC container would still used it when needed;
- lifetime of the service would be controlled by IoC container only, users of the interface won't be able to dispose the service instance (it's OK if it is injected "per dependency", but not OK if it is a "singleton").

pavel-leonenko (dismissed):

PR #541, opened 2024-02-07
Fixed issue when updated push token was not sent to backend. Reverted…

pavel-leonenko:
… disposing semaphore

### Issues Resolved

- Fixed issue when updated push token was not sent to backend

### API Changes

 None

### Platforms Affected

- Core (all platforms)

### Behavioral/Visual Changes

None

### Before/After Screenshots

Not applicable

### PR Checklist

### DIFF
--- a/Softeq.XToolkit.PushNotifications.Droid/Services/DroidPushNotificationsService.cs
+++ b/Softeq.XToolkit.PushNotifications.Droid/Services/DroidPushNotificationsService.cs
@@ -20,7 +20,7 @@ namespace Softeq.XToolkit.PushNotifications.Droid.Services
     ///     interfaces for Android platform. Handles all interactions with the platform, related to push notifications.
     /// </summary>
     public sealed class DroidPushNotificationsService :
-        IPushNotificationsService, IActivityLauncherDelegate, IDroidFirebaseMessagingHandler, IDisposable
+        IPushNotificationsService, IActivityLauncherDelegate, IDroidFirebaseMessagingHandler
     {
         private readonly IDroidPushNotificationsConsumer _pushNotificationsConsumer;
         private readonly ILogger _logger;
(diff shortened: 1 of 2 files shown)


Softeq.XToolkit.PushNotifications.Droid/Services/DroidPushNotificationsService.cs
dimysa, line 1: Why do you need to make changes to this file?

Softeq.XToolkit.PushNotifications/PushTokenSynchronizer.cs
```diff
@@ -130,14 +130,11 @@ await UnregisterFromRemotePushNotificationsAsync()
         private async Task OnRegisterSuccessInternalAsync(string token)
         {
             IsTokenRegisteredInSystem = true;
+            IsTokenSavedOnServer = false;
```
dimysa: I understand that `OnRegisterSuccessInternalAsync` is only called when the token has changed, but it would probably be more obvious to do `IsTokenSavedOnServer = false` in the `OnTokenChanged` method before calling the `OnRegisterSuccessInternalAsync` method

dimysa (changes requested):

Softeq.XToolkit.PushNotifications/PushTokenSynchronizer.cs
pavel-leonenko: I'm not sure about that. `IsTokenSavedOnServer` relates to the specific token which is stored in `_pushTokenStorageService`. Once we change `_pushTokenStorageService.PushToken`, we immediately mark that this specific token is not saved on server, so it's better to keep these two lines together.
`IsTokenRegisteredInSystem` is also set to true here, so it looks consistent for me.

dimysa (approved):

Issue #510, opened 2023-01-10
ViewModel lifecycle methods have different behavior on each platform


wcoder:
### Description

ViewModel lifecycle methods:

https://github.com/Softeq/XToolkit.WhiteLabel/blob/8af5c6aa542c72bfc21679dd00d1c177e482517b/Softeq.XToolkit.WhiteLabel/Mvvm/IViewModelBase.cs#L23-L33

https://github.com/Softeq/XToolkit.WhiteLabel/blob/8af5c6aa542c72bfc21679dd00d1c177e482517b/Softeq.XToolkit.WhiteLabel/Mvvm/ViewModelBase.cs#L33-L41

ViewModel lifecycle methods have different behavior on each platform:

### Expected Behavior

- iOS:
- Android: 

### Actual Behavior

- iOS:
- Android:

PR #569, opened 2024-12-02
Added possibility to conditionally notify ViewModel about appearing/disappearing

pavel-leonenko:
### Description

`ViewControllerBase` always calls `ViewModel.OnAppearing` method in `ViewWillAppear`, but sometimes it's needed to customize this logic. For example, we may need to skip calling this if the app is in background, and instead a `ViewController` would call `ViewModel.OnAppearing` only when the app is restored from background.

### API Changes

Added:
 - `protected virtual void NotifyViewModelAboutAppearing()`
 - `protected virtual void NotifyViewModelAboutDisappearing()`

### Platforms Affected

- iOS

### Behavioral/Visual Changes

None

### Before/After Screenshots

Not applicable

### PR Checklist


yauhenipakala:
ref: #510

dimysa (approved):

